PR #1, opened 2017-04-01
Modify interpreter to use completion values

cpcallen:
All existing tests pass.  Net change: -131 lines.

I could just merge the branch myself, but I figure you should have a look at it before I do, since this represents the biggest change yet to the architecture of the interpreter (compared to JS-interpreter).

### DIFF
--- a/server/interpreter/cval.go
+++ b/server/interpreter/cval.go
@@ -26,8 +26,7 @@ import (
 type cvalType int
 
 const (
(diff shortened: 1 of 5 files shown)


cpcallen:
Implemented ContinueStatement and added some more tests based on https://github.com/NeilFraser/JS-Interpreter/issues/93 (but modified, because I haven't implemented ForStatement yet).

PR #2, opened 2017-04-24
Introduce boxed primitives

cpcallen:
We were originally planning to exclude boxed primitives (e.g., new Number(42)) from our implementation, because they are not very useful and are kind of confusing for beginners.

Because we still want to be able to invoke methods on primitives, we therefore previously had the Value interface (representing any arbitrary JS value) include lots of internal methods (like Get, Set and HasOwnProperty) that are only really applicable to object types;  these were implemented on primitives in way that simulated the observed behaviour.  E.g.:

    “foo”.length         // == 3
    “foo”.bar = 42     // does nothing
    “foo”.length = 4  // error!

The ES5.1 spec actually says that in a MemberExpression where the left hand side is a primitive (i.e., “foo”.length) the interpreter should create a boxed primitive (== new String(“foo)) and use that as the actual left-hand-side.  Because this box is thrown away after the MemberExpression has finished evaluation, only methods inherited from {Boolean,Number,String}.prototype would be able to tell the difference: with a boxed primitive, they could write a property to the box and then read it back (but only until the method call ends, unless they save a reference to the box), while with our primitives-masquerading-as-objects the write would just silently fail.

That’s not really a big deal, since the primitive prototype methods specified in ES5.1 don’t ever need to write to the box, and if we add additional methods to the primitive prototypes we could anticipate this slight non-conformity.

It turns out, however, that introducing boxed primitives for internal use in MemberExpressions (even if we don’t let users create them manually using the ‘new’ keyword) lets us improve the code in the following ways:

- The various internal methods that only apply to objects can be separated out into a new Object interface (that itself still implements Value); these methods can then be removed from the primitive types.

- Because the primitives no longer have a Proto() method, we can more easily solve the problem of primitives finding their (non-primitive) prototype object.

Previously, primitives found their prototype by looking at package-global variables data.BooleanProto, data.NumberProto, etc.; this meant that prototypes were incorrectly shared between multiple Interpreter instances (i.e., a property added to a primitive prototype in one Interpreter would incorrectly be visible in a separate second Interpreter instance).  The possible solutions were:

- Have every primitive store a pointer to its prototype.  This is quite memory-inefficient, because it would increase the size of Booleans and Numbers by at leat 50% (assuming type tag, float64 and pointer to prototype are all 8 bytes each).

- Have a “raft” of primitive protos (an “environment” object, env) which could be passed in to the .Proto() method; Object.Proto(protos) would ignore this extra argument but {Boolean,Number,String}.Proto(protos) would return photos.{Boolean,Number,String}Proto as appropriate.  Unfortunately, since Prototypes is called from Get, HasProperty, and other methods it ends up being necessary to pass this environment object around all over the place even though it is rarely used.

- Use boxed primitives, and let MemberExpression (and lvalue) worry about finding the correct prototype when boxing a primitive.  This still necessitates keeping a raft of primitives in the interpreter object, and passing a reference to it to every stateFooBar.step() method, but (a) at least we don’t have to pass it to almost every internal method on every Value object, and (b) we were almost certainly eventually going to have to give the sate step functions a way of accessing their enclosing interpreter anyway.

### DIFF
--- a/server/interpreter/builtin_array.go
+++ b/server/interpreter/builtin_array.go
@@ -22,12 +22,12 @@ import (
 	"encoding/json"
 )
 
-func initBuiltinArray(sc *scope) {
-	// FIXME: should be Function:
-	var Array = data.NewObject(nil, data.ObjectProto)
+func initBuiltinArray(protos *data.Protos, sc *scope) {
+	// FIXME: should be Function?:
+	var Array = data.NewObject(nil, protos.FunctionProto)
 	sc.newVar("Array", Array)
 
-	Array.Set("prototype", data.ArrayProto)
+	Array.Set("prototype", protos.ArrayProto)
 
 	var params []*ast.Identifier
 	var body *ast.BlockStatement
@@ -40,8 +40,9 @@ func initBuiltinArray(sc *scope) {
 	if e != nil {
 		panic(e)
 	}
-	var push = newClosure(nil, sc, params, body)
-	data.ArrayProto.Set("push", push)
+	// FIXME: set owner:
+	var push = newClosure(nil, protos.FunctionProto, sc, params, body)
+	protos.ArrayProto.Set("push", push)
 }
 
 const pushPolyfillParams = `[{"type":"Identifier","start":32,"end":33,"name":"e"}]`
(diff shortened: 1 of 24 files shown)


cpcallen:
Closing this pull request because it has too much extraneous stuff in it.  The main change is complicated enough...

PR #4, opened 2017-04-24
Implement boxed primitives & autoboxing

cpcallen:
We were originally planning to exclude boxed primitives (e.g., `new Number(42)`) from our implementation, because they are not very useful and are kind of confusing for beginners.

Because we still want to be able to invoke methods on primitives, we therefore previously had the `data.Value` interface (representing any arbitrary JS value) include lots of internal methods (like `Get`, `Set` and `HasOwnProperty`) that are only really applicable to object types;  these were implemented on primitives in way that simulated the observed behaviour.  E.g.:

    “foo”.length      // == 3
    “foo”.bar = 42    // does nothing
    “foo”.length = 4  // error!

The ES5.1 spec actually says that in a MemberExpression (e.g., `foo.bar` or `foo[bar]`), where the left hand side is a primitive (i.e., `“foo”.length`) the interpreter should create a boxed primitive (`== new String(“foo")`) and use that as the actual left-hand-side.  Because this box is thrown away after the MemberExpression has finished evaluation, only methods inherited from {`Boolean`,`Number`,`String`}`.prototype` would be able to tell the difference: with a boxed primitive, they could write a property to the box and then read it back (but only until the method call ends, unless they save a reference to the box), while with our primitives-masquerading-as-objects the write would just silently fail.

(That’s not really a big deal, since the primitive prototype methods specified in ES5.1 don’t ever need to write to the box, and if we add additional methods to the primitive prototypes we could anticipate this slight non-conformity.)

It turns out, however, that introducing boxed primitives for internal use in MemberExpressions (even if we don’t let users create them manually using the `new` keyword) lets us improve the code in the following ways:

- The various internal methods that only apply to objects can be separated out into a new `Object` interface (that still implements `Value`); these methods can then be removed from the primitive types.

- Because the primitives no longer have a `Proto` method, we can more easily solve the problem of primitives finding their (non-primitive) prototype object.

In regards to this last point:

Previously, primitives found their prototype by looking at package-global variables `data.BooleanProto`, `data.NumberProto`, etc.; this meant that prototypes were incorrectly shared between multiple `Interpreter` instances (i.e., a property added to a primitive prototype in one `Interpreter` would incorrectly be visible in a separate second `Interpreter` instance).  This might not be a huge problem for us—we don’t at the moment envision having more than one interpreter—but it would be a nasty surprise to anyone else using the package.  The possible solutions were:

- Have every primitive store a pointer to its prototype.  This is quite memory-inefficient, because it would increase the size of `Booleans` and `Numbers` by at leat 50% (assuming the type tag, `float64` or `bool`, and pointer to prototype are all 8 bytes each).

- Have a “raft” of primitive prototypes (`protos`) which could be passed in to the `.Proto()` method; `Object.Proto(protos)` would ignore this extra argument but {`Boolean`,`Number`,`String`}`.Proto(protos)` would return `photos.`{`Boolean`,`Number`,`String`}`Proto` as appropriate.  Unfortunately, since `Proto` is called from `Get`, `HasProperty`, and other methods it ends up being necessary to pass this raft object around all over the place even though it is rarely used.

- Use boxed primitives, and let MemberExpression (and lvalue) worry about finding the correct prototype when boxing a primitive.  This still necessitates keeping a raft of primitives in the `Interpreter` object, and passing a reference to it to every `state`*FooBar*`.step()` method, but (a) at least we don’t have to pass it to almost every internal method on every `Value` object, and (b) we were almost certainly eventually going to have to give the sate `step` methods a way of accessing their enclosing `Interpreter` anyway.


server/interpreter/data/object.go
```diff
@@ -16,11 +16,48 @@
...
+	/// Any Object is a valid Value.
+	Value
+
+	// Proto returns the prototype (parent) object for this object.
+	// N.B. this is object.__proto__, not Constructor.prototype!
+	Proto() Object
+
+	// Get returns the current value of the given property or an
+	// ErrorMsg if that was not possible.
+	Get(name string) (Value, *ErrorMsg)
```
tombergan: Returning an error (rather than a custom error type, like *ErrorMsg) is more idiomatic Go. I realize structured errors can be useful, but usually this is done by casting to a type from error rather than returning that type directly. Also, the ErrorMsg type doesn't look like it has particularly useful fields (do you often extract these fields individually?).
https://github.com/NeilFraser/CodeCity/blob/master/server/interpreter/data/error.go

server/interpreter/data/object.go
```diff
@@ -16,11 +16,48 @@
...
-// Object represents typical JavaScript objects with (optional)
-// prototype, properties, etc.
-type Object struct {
+// Object represents any JavaScript value (primitive, object, etc.).
+type Object interface {
+	/// Any Object is a valid Value.
+	Value
+
+	// Proto returns the prototype (parent) object for this object.
+	// N.B. this is object.__proto__, not Constructor.prototype!
```
tombergan: Technically it's Object.getPrototypeOf() not object.__proto__, which can be overwritten. I think?

server/interpreter/data/owner.go
```diff
@@ -18,12 +18,12 @@ package data
 
 // An Owner is an object that can own other objects and properties.
```
tombergan: I am moderately fluent in JavaScript, but I can't figure out what this is from the description. Should I be able to?

server/interpreter/data/protos.go
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package data
+
+// Protos is a struct containing prototypes for the various Object
```
tombergan: Is `Builtins` a better name?

server/interpreter/data/boxes_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}{
+		{"", 0},
+		{"Hello, World!", 13},
+		{"కోడ్ సిటీ", 9},
+		{"𝌆", 2},
+	}
+	for _, c := range tests {
+		bstr := NewBoxedString(nil, protos.StringProto, String(c.in))
+		if l, _ := bstr.Get("length"); l != Number(c.expected) {
+			t.Errorf("new String(%#v).length == %d (expected %d)",
```
tombergan: line break not needed

server/interpreter/data/boxes_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if !s.HasOwnProperty("length") {
+		t.Errorf("%#v.HasOwnProperty(\"length\") == false", s)
+	}
+}
+
+func TestBoxedStringHasProperty(t *testing.T) {
+	var s = NewBoxedString(nil, protos.StringProto, String("foo"))
+
+	if s.HasProperty("foo") {
+		t.Errorf("%#v.HasProperty(\"foo\") == true", s)
```
tombergan: You can use raw string literals (backtick) to avoid needing to escape quotes

server/interpreter/data/boxes.go
```diff
@@ -0,0 +1,237 @@
...
+	bstr.init(owner, proto)
+	bstr.f = true
+	bstr.value = value
+	return bstr
+}
+
+/********************************************************************/
+
+// Coerce coerces its first argument into an object.
+func Coerce(value Value, owner *Owner, protos *Protos) Object {
```
tombergan: This is probably fine, but since this depends on a set of builtins, it might work better to have this be a method of Interpreter (or some other runtime-global object) so you don't have to thread protos through every function.

server/interpreter/state.go
```diff
@@ -877,8 +880,10 @@ func (st *stateFunctionExpression) init(node *ast.FunctionExpression) {
 	st.body = node.Body
 }
 
-func (st *stateFunctionExpression) step(cv *cval) (state, *cval) {
-	return st.parent, pval(newClosure(nil, st.scope, st.params, st.body))
+func (st *stateFunctionExpression) step(intrp *Interpreter, cv *cval) (state, *cval) {
+	// FIXME: set owner:
+	return st.parent, pval(newClosure(nil, intrp.protos.FunctionProto,
+		st.scope, st.params, st.body))
```
tombergan: no line break

server/interpreter/data/boxes.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Set on BoxedString will reject attempts to set .length or numeric
+// character indicies, and otherwise delegates to the embedded object.
+//
+// BUG(cpcallen): character indicides not implemented.
+func (bstr *BoxedString) Set(key string, value Value) *ErrorMsg {
+	if key == "length" {
+		return &ErrorMsg{"TypeError",
+			fmt.Sprintf("Cannot assign to read only property '%s' of %s",
```
tombergan: no line break

server/interpreter/data/boxes.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Delete on BoxedString will refuse to delete "length" and numeric
+// character indicies, and otherwise defers to the embedded object.
+//
+// BUG(cpcallen): character indicides not implemented.
+func (bstr *BoxedString) Delete(key string) *ErrorMsg {
+	if key == "length" {
+		return &ErrorMsg{"TypeError",
+			fmt.Sprintf("Cannot delete property '%s' of %s",
```
tombergan: no line break

server/interpreter/data/object.go
cpcallen: You are 100% correct.  I have rewritten to make it more clear.

server/interpreter/data/owner.go
cpcallen: Owner is a non-standard extension, so I think it warrants a slightly more detailed explanation which I have now provided.

server/interpreter/data/protos.go
cpcallen: To me, Builtins would include (for example) Object, not just Object.prototype.  Protos carries only the prototypes, not any of the constructors (or libraries, like Math).

server/interpreter/data/boxes_test.go
cpcallen: Oh, thanks!

server/interpreter/state.go
cpcallen: O.K.

server/interpreter/data/boxes.go
cpcallen: O.K.

server/interpreter/data/boxes.go
cpcallen: O.K.

server/interpreter/data/boxes.go
cpcallen: An excellent suggestion.  I've moved this to interpreter.go, given it an *Interpreter receiver, and renamed it to toObject (to correspond with the name it is given in the ES spec).

I still have to pass a pointer to the Interpreter object to each stateWhatever.step() call, which is a little ugly.  It would be more elegant if I could make the various states into inner classes of Interpreter, so they could get implicit (lexical) access to the interpreter's members/methods, but Go doesn't have inner classes.  I'm open to suggestions if there is a more idomatically Go alternative to the one I've picked.

server/interpreter/data/object.go
cpcallen: The intention is that ErrorMsg objects returned by the data package will be turned into a new JS `Error` object of the appropriate type by the caller, which would then be thrown.  However:

- I haven't actually written the code to do that yet.
- It might be better to replace `Name string` with an enum-valued field, since there are only about a half-dozen different error types (`SyntaxError`, `TypeError`, `ReferenceError` etc.)
- It would be nice to replace the error return with a "throw" flag and just create and return the relevant JS `Error` object as the first return value, but that requires a prototype and I want to avoid having to pass a Protos object to every single function/method that might need to throw an error.
- Fulfilling the Go `error` interface was mainly so that I could `panic()` these errors as a temporary measure until I implement the code to cause them to be turned into `Error`s and and throw.

I would be happy to prepare an improved version of this as part of this PR, but I invite your advice before I do so.

cpcallen:
Can you provide any further advice on line breaking?  From my previous stint at Google I'm pretty used to the absolute 80-character limit (except 100 in Java, because Java).  I don't mind having some flexibility here, but some guidance would be helpful.

server/interpreter/data/object.go
tombergan: My suggestion would be one of the following:

- If ErrorMsg is really only intended to be used where JavaScript would throw an exception, rename the type to Exception.

- Return an ordinary error interface and add another interface that can extract an exception from an error. The caller would typecast to that interface to check if the error should throw an exception (`err.(Exception)`).

tombergan:
Re: line breaks. I thought I had left comments somewhere on the CL, but can't find them. Generally speaking, Go has no line length limit and line breaks are rare. I'd say they mostly show up in only two cases:

- if statements with very long conditions
- printf calls with many arguments

That is generally speaking -- there is no strict rule.

server/interpreter/data/object.go
cpcallen: I've gone with the first of these (since I was going to go in that direction anyway), except that I have picked the name "NativeError" rather than "Exception", to align more closely with the ECMA262 specification.  The changes are in [a subsidiary pull request](https://github.com/NeilFraser/CodeCity/pull/5) for clarity.

cpcallen:
I've removed some extraneous line breaks as I've gone along and/or where I knew there were a bunch of them.

Let me know if you have any further comments on this PR or [the subsidiary one](https://github.com/NeilFraser/CodeCity/pull/5).

tombergan:
lgtm

PR #5, opened 2017-05-02
Replace ErrorMsg with NativeError; actually throw Errors

cpcallen:
NativeError doesn’t implement the go Error interface, to make it clear
this is a template for JS Error objects rather than a structured Go
error type.

Add Interpreter.newError (to construct JS Error objects) and
Interpreter.throw (to throw them); use these where previously we
panic()ed a ErrorMsg.


tombergan:
lgtm

PR #15, opened 2017-06-22
Add tests

cpcallen:
Tests can be run from anywhere under server/ by typing ‘npm test’, or manually from server proper by running ’node tests/run.js’.

Many tests fail.

The testing framework is extremely simple (it doesn’t even keep track of how many pass/fail).

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -23,11 +23,13 @@
  */
 'use strict';
 
+var acorn = require("../third_party/acorn/acorn");
+
 /**
  * Create a new interpreter.
  * @constructor
  */
-var Interpreter = function() {
+function Interpreter() {
   this.paused_ = false;
   // Unique identifier for native functions.  Used in serialization.
   this.functionCounter_ = 0;
@@ -3383,3 +3385,5 @@ Interpreter.prototype['stepWithStatement'] = function() {
 
 Interpreter.prototype['stepWhileStatement'] =
     Interpreter.prototype['stepDoWhileStatement'];
+
+module.exports = Interpreter;
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -23,13 +23,13 @@
  */
 'use strict';
 
-var acorn = require("../third_party/acorn/acorn");
+var acorn;
 
 /**
  * Create a new interpreter.
  * @constructor
  */
-function Interpreter() {
+var Interpreter = function () {
   this.paused_ = false;
   // Unique identifier for native functions.  Used in serialization.
   this.functionCounter_ = 0;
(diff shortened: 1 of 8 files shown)


server/interpreter.js
```diff
@@ -3386,4 +3386,7 @@ Interpreter.prototype['stepWithStatement'] = function() {
 Interpreter.prototype['stepWhileStatement'] =
     Interpreter.prototype['stepDoWhileStatement'];
 
-module.exports = Interpreter;
+if (typeof module !== 'undefined') { // Node.js
```
NeilFraser: Only need != here.  There are no tricky values at play.

server/tests/interpreter_test.js
```diff
@@ -0,0 +1,38 @@
...
+  if (!('expected' in tc)) {
+    console.log('SKIP\t%s', tc.name);
+    continue;
+  }
+  
+  var interpreter = new Interpreter()
+  interpreter.appendCode(autoexec);
+  interpreter.run();
+
+  var err = undefined;
```
NeilFraser: No need for ' = undefined'.

server/interpreter.js
```diff
@@ -23,11 +23,13 @@
...
 'use strict';
 
+var acorn;
+
 /**
  * Create a new interpreter.
  * @constructor
  */
-var Interpreter = function() {
+var Interpreter = function () {
```
NeilFraser: No space here.

cpcallen: I will comply, but see https://engdoc.corp.google.com/eng/doc/devguide/js/styleguide.md?cl=head#formatting-whitespace

server/interpreter.js
cpcallen: Wontfix.  See https://groups.google.com/a/google.com/forum/#!search/javascript$20equality%7Csort:date/javascript-style/wSOzSX6BCKw/6gVzHM2-DgAJ for the most recent discussion I can find on this, but in summary:

* `==` was previously (ca. 2007) mandated by the styleguide but has not been for some time (i.e., in at least the current and most recent previous editions).
* Most of those arguing in favour of `==` by default are:
   * now only making that argument in only respect to large existing google3 code bases where I agree that it does offer some readability benefits (`interpreter.js` is large but not that large yet...), and
   * assuming the code is going  to be compiled (ours will not be, AFAIK) and so will be type-checked.
* Even many newer google3 JS projects now mandate `===`.
* Outside of Google`===` is effectively the industry standard.

We should be using it in-core for pedagogical reasons, and in the server implementation because we are an open-source project and should be facilitating external contributions.

server/tests/interpreter_test.js
cpcallen: Yes there is, because 'var' is not block scoped, and needs to be undefined at this point in the loop.

Changed to let, along with other styleguide improvements (this file only)

PR #17, opened 2017-06-23
Depseudo

NeilFraser:


server/interpreter.js
```diff
@@ -2223,8 +2007,7 @@ Interpreter.prototype.populateScope_ = function(node, scope) {
 
 /**
  * Remove start and end values from AST, or set start and end values to a
- * constant value.  Used to remove highlighting from polyfills and to set
- * highlighting in an eval to cover the entire eval expression.
+ * constant value.  Used to remove highlighting from polyfills.
```
cpcallen: Perhaps this comment should still mention eval expressions.

server/interpreter.js
```diff
@@ -2880,16 +2650,7 @@ Interpreter.prototype['stepForInStatement'] = function() {
...
-            state.visited_.add(prop);
-            state.name_ = prop;
-            state.visited_.add(prop);
-            break done;
-          }
-        }
-      } else {
+      if (state.object_.isObject) {
         for (var prop in state.object_.properties) {
           if (state.visited_.has(prop)) {
```
cpcallen: This test is inverted, which explains why all the for..in tests are failing at the moment.

server/interpreter.js
```diff
@@ -2900,6 +2661,15 @@ Interpreter.prototype['stepForInStatement'] = function() {
             }
           }
         }
+      } else {
+        for (var prop in state.object_) {
+          if (state.visited_.has(prop)) {
```
cpcallen: Test inverted.

server/interpreter.js
```diff
@@ -2900,6 +2661,15 @@ Interpreter.prototype['stepForInStatement'] = function() {
             }
           }
         }
+      } else {
+        for (var prop in state.object_) {
```
cpcallen: Is this case to deal with the case of `for (x in "foo")`?  If so, it is wrong, because it ends up iterating over the properties of `String.prototype` instead of `Interpreter.STRING.prototype` (and similarly for other primitives).

(This is one case where "destructive" testing—i.e., the sort you wouldn't want to do on a running production core—can be very useful.)

server/interpreter.js
```diff
@@ -3303,18 +3064,17 @@ Interpreter.prototype['stepUnaryExpression'] = function() {
...
-                          name + '\' of \'' + obj + '\'');
-      return;
+      this.throwException(this.TYPE_ERROR, "Cannot delete property '" +
+                          name + "' of '" + obj + "'");
     }
-  } else if (node['operator'] == 'typeof') {
-    value = value.type;
-  } else if (node['operator'] == 'void') {
+  } else if (node['operator'] === 'typeof') {
+    value = value.node ? 'function' : typeof value;
```
cpcallen: This should identify user functions (and therefore polyfills), but will it correctly identify native functions?  Perhaps you should check `.isFunction`?

server/interpreter.js
NeilFraser: Reverted.

server/interpreter.js
NeilFraser: Fixed.

server/interpreter.js
NeilFraser: Fixed.

server/interpreter.js
NeilFraser: Fixed.

server/interpreter.js
NeilFraser: Fixed.

server/interpreter.js
```diff
@@ -171,6 +170,7 @@ Interpreter.prototype.run = function() {
  */
 Interpreter.prototype.initGlobalScope = function(scope) {
   // Initialize uneditable global properties.
+  this.addVariableToScope(scope, 'this', undefined, true);
```
cpcallen: Oh, nice.

server/interpreter.js
```diff
@@ -479,21 +469,19 @@ Interpreter.prototype.initObject = function(scope) {
       Interpreter.NONENUMERABLE_DESCRIPTOR);
 
   wrapper = function(obj, prop) {
-    prop = (prop || thisInterpreter.UNDEFINED).toString();
+    throwIfNullUndefined(obj);
+    prop += '';
```
cpcallen: I guess we need to balance that against the fact that `+ ''` calls `valueOf` in preference to `toString`, as you have recently discovered.  That might explain the speed difference: producing the wrong answer might take less time / do less allocating...

server/interpreter.js
```diff
@@ -532,50 +517,46 @@ Interpreter.prototype.initObject = function(scope) {
 
   // Instance methods on Object.
   wrapper = function() {
-    return thisInterpreter.createPrimitive(this.toString());
+    return this.toString();
```
cpcallen: And this is now fixed (pending changes in your local client).

PR #18, opened 2017-06-26
Improve test harness; keep count of pass/fail/crash/skip

cpcallen:


cpcallen:
Verbal LGTM from @NeilFraser.

PR #19, opened 2017-06-27
Beginnings of a core: $.physical

cpcallen:
Tests are more of a demo than actual unit tests so far.

### DIFF
--- a/core/core.js
+++ b/core/core.js
@@ -1,0 +1,164 @@
+'use strict'; // For testing with node.js
+
+// Global scope declarations:
+var $ = Object.create(Object.prototype);
+
+(function() {
+  // Core build options:
+
+  // Overwrite existing?:
+  var force = false;
+
+  /**
+   * Add a property to a core object - but only overwrite existing
+   * properties if force is true.
+   * @param {Object} obj The object to which to add a property.
+   * @param {string} key The property key to add.
+   * @param {*} value The property key to add.
+   */
+  function make(obj, key, value) {
+    if (force || typeof obj[key] === 'undefined') {
+      obj[key] = value;
+    }
+  }
+
+  /*******************************************************************
+   * $.object
+   */
+  make($, 'object', Object.prototype);
+
+  /*******************************************************************
+   * $.physical
+   */
+  make($, 'physical', Object.create($.object, {
+    name: {value: 'Physical object prototype',
+	   writable: true, enumerable: true, configurable: true},
+    location: {value: null,
+	       writable: true, enumerable: true, configurable: true},
+    contents_: {value: undefined, // wil be vetted into existance
+		writable: true, enumerable: true, configurable: true},
+  }));
+
+  make($.physical, 'contents', function() {
+    // Return the contents of this object.  This is for VR purposes,
+    // and descendents of $.physical can override it to 'hide' certain
+    // objects from their contents.
+    $.physical_vet(this);
+    return this.contents_;
+  });
+       
+  make($.physical, 'moveto', function(dest) {
+    // Move this physical object to dest
+    //
+    // This is based loosely on moo.ca's #3:moveto and #102:bf_move
+    // Vet obj, src and dest (also does type check):
+    $.physical_vet(this);
+    var src = this.location;
+    src === null || $.physical_vet(src);
+    dest === null || $.physical_vet(dest);
+
+    // FIXME: permission checks go here.
+    // Do moveable and accept checks:
+    if (!this.moveable(dest)) {
+      var destname = (dest === null ? 'null' : dest.name);
+      throw Error(this.name + ' declined to move to ' + destname);
+    }
+    if (dest !== null && !dest.accept(this)) {
+      throw Error(dest.name + ' refused ' + this.name);
+    }
+
+    // Check proposed containment hierarchy for cycles.  We do this
+    // after the movable and accept checks because we don't want those
+    // checks in case they muck arounnd with the containment
+    // hierarchy.
+    for(var loc = dest; loc !== null; loc = loc.location) {
+      if (loc === this) {
+	throw Error("Can't put " + this.name + " inside itself.");
+      }
+    }
+
+    // Do move:
+    var i;
+    if ($.physical.isPrototypeOf(src)) {
+      while ((i = src.contents_.indexOf(this)) >= 0) {
+	src.contents_.splice(i, 1);
+      }
+    }
+    this.location = dest;
+    dest === null || dest.contents_.push(this);
+
+    // FIXME: call exitfunc
+    // FIXME: call enterfunc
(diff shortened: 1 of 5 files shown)


core/core.js
```diff
@@ -0,0 +1,164 @@
+'use strict'; // For testing with node.js
+
+// Global scope declarations:
+var $ = Object.create(Object.prototype);
```
NeilFraser: var $ = {};

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+
+  /*******************************************************************
+   * $.physical
+   */
+  make($, 'physical', Object.create($.object, {
+    name: {value: 'Physical object prototype',
+	   writable: true, enumerable: true, configurable: true},
+    location: {value: null,
+	       writable: true, enumerable: true, configurable: true},
+    contents_: {value: undefined, // wil be vetted into existance
```
NeilFraser: will

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+   * $.object
+   */
+  make($, 'object', Object.prototype);
+
+  /*******************************************************************
+   * $.physical
+   */
+  make($, 'physical', Object.create($.object, {
+    name: {value: 'Physical object prototype',
+	   writable: true, enumerable: true, configurable: true},
```
NeilFraser: In the case of true/true/true, why not just assign the property normally?
That said, should these be enumerable?

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+	       writable: true, enumerable: true, configurable: true},
+    contents_: {value: undefined, // wil be vetted into existance
+		writable: true, enumerable: true, configurable: true},
+  }));
+
+  make($.physical, 'contents', function() {
+    // Return the contents of this object.  This is for VR purposes,
+    // and descendents of $.physical can override it to 'hide' certain
+    // objects from their contents.
+    $.physical_vet(this);
```
NeilFraser: The Google Style guide would ask that this be called physicalVet.

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+      throw Error(dest.name + ' refused ' + this.name);
+    }
+
+    // Check proposed containment hierarchy for cycles.  We do this
+    // after the movable and accept checks because we don't want those
+    // checks in case they muck arounnd with the containment
+    // hierarchy.
+    for(var loc = dest; loc !== null; loc = loc.location) {
+      if (loc === this) {
+	throw Error("Can't put " + this.name + " inside itself.");
```
NeilFraser: Indent one more space.

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+    // Vet obj, src and dest (also does type check):
+    $.physical_vet(this);
+    var src = this.location;
+    src === null || $.physical_vet(src);
+    dest === null || $.physical_vet(dest);
+
+    // FIXME: permission checks go here.
+    // Do moveable and accept checks:
+    if (!this.moveable(dest)) {
+      var destname = (dest === null ? 'null' : dest.name);
```
NeilFraser: Likewise, destName

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+      throw Error(this.name + ' declined to move to ' + destname);
+    }
+    if (dest !== null && !dest.accept(this)) {
```
NeilFraser: How about just !dest, instead of a null-only check.  Or: !dest || !dest.accept || ...

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+    // function.  This function (and any overrides) MUST NOT cause any
+    // observable action to occur.
+    return false;
+  });
+
+  make($.physical, 'contains', function(obj) {
+    // Returns true iff obj is located in this.
+    $.physical_vet(this);
+    $.physical_vet(obj);
+    for (var loc = obj.location; loc !== null; loc = loc.location) {
```
NeilFraser: Drop the !== null?

core/core.js
```diff
@@ -0,0 +1,164 @@
...
+  // FIXME: should be a non-overridable method on $.physical:
+  make($, 'physical_vet', function(obj) {
+    // Verify the integrity of a $.physical object.
+    if (!$.physical.isPrototypeOf(obj)) {
+      throw TypeError('Not a $.physical object');
+    }
+    // They can only be located in another $.physical object - and
+    // that object must obj in its contents:
+    if (!$.physical.isPrototypeOf(obj.location) ||
+	obj.contents_.indexOf(obj) == -1) {
```
NeilFraser: ===

core/core.js
```diff
@@ -0,0 +1,164 @@
+'use strict'; // For testing with node.js
```
NeilFraser: Add an Apache header.

cpcallen: Done.

core/core.js
cpcallen: I was attempting to standardise on Object.create for all object creation, but perhaps in this case that is not really necessary.

core/core.js
cpcallen: Done.

core/core.js
cpcallen: I am thinking of this as a placeholder for $.physical.vet, awaiting non-overridable properties.

core/core.js
cpcallen: Done.

core/core.js
cpcallen: The latter is too long to fit on one line, unfortunately...

Actually, we should probably think about to what extent we want to encourage || and && as conditionals.  They are very good for conciseness, which is probably why Go prevents them from being used in that way.  I don't know what standard practice is in JS; the styleguide says nothing about it AFAICT.

core/core.js
cpcallen: The indentation is actually correct, but Emacs and GitHub have conspired to convert larger indents to use tabs (fewer bytes, doncha know!) and then have only lines starting with spaces be moved one character to the right by the leading +.

I thought I had got Emacs successfully configured to use Google's official Emacs config for JS (which presumably should not use tabs, since the styleguide forbids them) but apparently either my setup or the official configuration is broken.

core/core.js
cpcallen: Done.

core/core.js
cpcallen: As in the above cases, I think the `=== null` check makes it clear that we are looking for objects whose location is 'nowhere', rather than merely a somehow invalid location.

cpcallen:
Fooey: I forot that I mustn't rebase my local branch while GitHub PR exists for it, and now I've made a mess of things.  Bother.

cpcallen:
O.K., `git rebase --interactive` and `git push --force` FTW.  All fixed now.

PR #20, opened 2017-06-28
Refs should always be tuples.

NeilFraser:


server/interpreter.js
```diff
@@ -2744,12 +2747,12 @@ Interpreter.prototype['stepMemberExpression'] = function() {
     stack.push({node: node['property'], components: !node['computed']});
```
cpcallen: As noted: if `!computed`, `node.property` will be of type `Identifer` and you can just set `state.propName = node.property.name`.

server/interpreter.js
```diff
@@ -2306,9 +2310,8 @@ Interpreter.prototype['stepCallExpression'] = function() {
   }
   if (!state.doneExec_) {
     state.doneExec_ = true;
-    if (state.func_ === null || state.func_ === undefined) {
-      this.throwException(this.TYPE_ERROR,
-          (typeof state.func_) + ' is not a function');
+    if (!state.func_ || !this.isa(state.func_, this.FUNCTION)) {
```
cpcallen: This check is wrong.  Probably best just to just check we have an object here, and throw 'not a function' below if you can't figure out how to call it.

server/interpreter.js
```diff
@@ -2741,19 +2741,30 @@ Interpreter.prototype['stepMemberExpression'] = function() {
...
+  if (!state.doneProperty_) {
+    state.object_ = state.value;
+    // obj[foo] -- Compute value of 'foo'.
+    state.doneProperty_ = true;
+    stack.push({node: node['property']});
+    return;
+  }
+  stack.pop();
+  var propName = state.value;
+  if (Array.isArray(propName)) {
```
cpcallen: Can't happen, surely?

server/interpreter.js
```diff
@@ -2741,19 +2741,30 @@ Interpreter.prototype['stepMemberExpression'] = function() {
...
+    propName = propName[1];
   }
+  stack[stack.length - 1].value = state.components ?
```
cpcallen: Don't repeat yourself:

```JS
  var propName;
  if (!node['computed']) {
    // obj.foo -- Just access 'foo' directly.
    propName = node['property']['name'];
  } else if (!state.doneProperty_) {
    state.object_ = state.value;
    // obj[foo] -- Compute value of 'foo'.
    state.doneProperty_ = true;
    stack.push({node: node['property']});
    return;
  } else {
    propName = state.value;
  }
  stack.pop();
  stack[stack.length - 1].value = state.components ?
      [state.object_, propName] : this.getProperty(state.object_, propName);
}
```

PR #31, opened 2017-07-01
Link the scope to every state object.

NeilFraser:
Improves performance of getScope, because looking up a non-existent
property is sometimes expensive.  4800ms -> 4450ms

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1793,12 +1793,11 @@ Interpreter.prototype.setNativeFunctionPrototype =
  * @return {!Interpreter.Scope} Current scope dictionary.
  */
 Interpreter.prototype.getScope = function() {
-  for (var i = this.stateStack.length - 1; i >= 0; i--) {
-    if (this.stateStack[i].scope) {
-      return this.stateStack[i].scope;
-    }
+  var scope = this.stateStack[this.stateStack.length - 1].scope;
+  if (!scope) {
+    throw Error('No scope found.');
   }
-  throw Error('No scope found.');
+  return scope;
 };
 
 /**
@@ -2048,6 +2047,19 @@ Interpreter.prototype.executeException = function(error) {
   throw realError;
 };
 
+/**
+ * Create and push a new state onto the statestack.
+ * @param {!Object} node AST node for the state.
+ * @return {!Object} New state.
+ */
+Interpreter.prototype.pushNode_ = function(node) {
+  var state = {
+    node: node,
+    scope: this.stateStack[this.stateStack.length - 1].scope
+  };
+  this.stateStack.push(state);
+  return state;
+};
 
 ///////////////////////////////////////////////////////////////////////////////
 // Functions to handle each node type.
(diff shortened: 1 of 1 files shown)


server/interpreter.js
```diff
@@ -2048,6 +2047,19 @@ Interpreter.prototype.executeException = function(error) {
   throw realError;
 };
 
+/**
+ * Create and push a new state onto the statestack.
+ * @param {!Object} node AST node for the state.
+ * @return {!Object} New state.
+ */
+Interpreter.prototype.pushNode_ = function(node) {
```
cpcallen: Is there a reason this is a method on `Interpreter.prototype`, rather than on `stateStack`?

Is there a reason not to make scope an optional parameter here?

server/interpreter.js
```diff
@@ -2085,7 +2097,7 @@ Interpreter.prototype['stepAssignmentExpression'] = function() {
   var node = state.node;
   if (!state.doneLeft_) {
     state.doneLeft_ = true;
-    stack.push({node: node['left'], components: true});
+    this.pushNode_(node['left']).components = true;
```
cpcallen: This is not very readable.  Perhaps `compenents` *and* `scope` should be optional parameters to `pushNode_`?

server/interpreter.js
NeilFraser: `stateStack` is a plain array.  To put it there means one of:

* Create the function elsewhere, and copy it onto the array each time it is created.
* Attempt to subclass Array.
* Create a wrapper object that contains the function, and an array, and an initializer.

Each of these seems to increase complexity.  Not sure what you have in mind.

server/interpreter.js
NeilFraser: `components`, `scope`, `throwValue`, and who knows what in the future.  I'm not sure that `this.pushNode_(node['foo'], null, null, value);` is more readable than `this.pushNode_(node['foo']).throwValue = value;`.

server/interpreter.js
cpcallen: Just that in `this.stateStack.pushNode_(node)` is both more concise and clearer than `this.pushNode_(this.StateStack, node)`.

I'd go with option #1.

server/interpreter.js
cpcallen: What about `…(…, {components: true, scope: s})`?

server/interpreter.js
NeilFraser: That would mean that `this` in that function will refer to something other than where the function is written.  It will be all alone with no other functions behaving like this.  The function will be well-hidden from IDEs and programmers alike, discoverable only by tracing the interpreter's initialization code.  I think this is the first time I've seen monkey patching suggested for improving clarity.

Also, the signature is just `this.pushNode_(node);`, stateStack is never mentioned in the call.

server/interpreter.js
NeilFraser: The goal of this PR is to speed up the interpreter.  Adding a for-in loop inside the interpreter's second-most called function adds about 150ms to the test.  The PR is still a net improvement, but it's a pretty steep cost.

Alternatively, we could spilt the lines in two:
`var node = this.pushNode_(node['foo']);`
`node.throwValue = value;`

cpcallen: Not sure why you'd want a for-in loop.  See #32.

server/interpreter.js
cpcallen: Good point about the signature.

PR #32, opened 2017-07-02
Modify .pushNode_() to take named parameters

cpcallen:
I was thinking more like this.

(It’s a pity that scope is one of the extra arguments; it would be cleaner otherwise.)


NeilFraser:
No longer needed.

PR #35, opened 2017-07-04
Simplify operators.

NeilFraser:


server/interpreter.js
```diff
@@ -3071,7 +3045,7 @@ Interpreter.prototype['stepSwitchStatement'] = function() {
         this.pushNode_(switchCase['test']);
         return;
       }
-      if (state.matched_ || this.comp(state.value, state.switchValue_) === 0) {
+      if (state.matched_ || state.value === state.switchValue_) {
```
cpcallen: Looks like this was actually wrong before.

server/interpreter.js
```diff
@@ -2274,48 +2242,48 @@ Interpreter.prototype['stepAssignmentExpression'] = function() {
...
-    } else {
+  var value = leftSide;
+  switch (node['operator']) {
+    case '=':
+      value = rightSide;
+      break;
+    case '+=':
+      value += rightSide;
+      break;
+    case '-=':
```
cpcallen: In this case I would be totally O.K. with ignoring the style guide and formatting these as:

```JS
  switch (node['operator']) {
    case '=':   value =   rightSide; break;
    case '+=':  value +=  rightSide; break;
```

I actually think it might actually be preferable because it would make it really obvious if there were any missing break statements, mismatched operators, variable misspellings, etc.

PR #37, opened 2017-07-04
Closure type error fixes with a side helping of constructor fixes and toString refactoring

cpcallen:
- Fixed lots of type errors (and several real bugs).
- Made sure Function, Array, Date, RegExp, and *Error constructors actually construct a value of the correct type (and not just with the correct prototype).
- Split up Interpreter.Object.prototype.toString and moved the bits to where they belong.
- Always use String() for string conversion.
- Added many tests.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1077,7 +1077,7 @@ Interpreter.prototype.initRegExp = function(scope) {
     if (!(this instanceof Interpreter.RegExp) ||
         !(this.regexp instanceof RegExp)) {
       thisInterpreter.throwException(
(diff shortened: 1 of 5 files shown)


server/interpreter.js
```diff
@@ -390,6 +371,9 @@ Interpreter.prototype.initObject = function(scope) {
   };
 
   // Static methods on Object.
+  this.setProperty(ObjectConst, 'is', this.createNativeFunction(Object.is),
```
cpcallen: As is `Object.setPrototypeOf()`…  And IIRC there was at least one feature we need that turned out to be ES7!

PR #38, opened 2017-07-05
Add scope_reference symbol.  Clarify reference vs value variables.

NeilFraser:


server/interpreter.js
```diff
@@ -104,6 +104,12 @@ Interpreter.READONLY_NONENUMERABLE_DESCRIPTOR = {
  */
 Interpreter.STEP_ERROR = {};
 
+/**
+ * Unique symbol for indicating that a reference is a variable on the scope,
```
cpcallen: Maybe "sentinel value" instead of "symbol" here (and above), least we confuse ES6 programmers (or our future selves).

PR #39, opened 2017-07-05
Fix Number.prototype.toString

NeilFraser:


server/tests/interpreter_test.js
```diff
@@ -606,3 +606,22 @@ exports.testAeca = function(t) {
     runTest(t, 'ASECA: ' + tc[0], src, tc[2]);
   }
 };
+
+exports.testNumber = function(t) {
```
cpcallen: It would probably have been sufficient to add a single test into testcases.js, but I won't complain about you writing *more* tests than are needed.

PR #40, opened 2017-07-05
Make Interpreter.Object and friends into non-static inner classes

cpcallen:
Now their methods can access the enclosing interpreter instance.


cpcallen:
I have had a suggestion about a slightly better way to fix the type warnings, so I will withdraw this for now.

PR #41, opened 2017-07-06
Make Interpreter.Object and friends into (non-static) inner classes of Interpreter

cpcallen:
The type error fixes in 3d513c7 are probably just temporary until I can figure out how to use @interface / @implements correctly.  The way I've done it here is not very nice (but does fix the warnings!)

The main thing that requires consideration is the refactor in 26585ca.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -30,6 +30,7 @@ var acorn;
  * @constructor
  */
 var Interpreter = function() {
+  this.installTypes();
   this.paused_ = false;
   // Unique identifier for native functions.  Used in serialization.
   this.functionCounter_ = 0;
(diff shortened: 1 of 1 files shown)


server/interpreter.js
```diff
@@ -2246,6 +1979,393 @@ Interpreter.prototype.pushNode_ = function(node) {
 };
 
 ///////////////////////////////////////////////////////////////////////////////
+// Types representing JS objects
+///////////////////////////////////////////////////////////////////////////////
+
+// This is a bunch of boilerplate that serves two purposes:
+// 
+// * First, by declaring these types as if they were on
+//   Interpreter.prototype we can get the closure compiler to type
```
NeilFraser: Closure Compiler

server/interpreter.js
```diff
@@ -2246,6 +1979,393 @@ Interpreter.prototype.pushNode_ = function(node) {
...
+//   check use of them for us.
+//
+// * Second, for whatever reason these declarations seem to create a
+//   small performance improvement.
```
NeilFraser: ... in V8.

server/interpreter.js

server/interpreter.js

PR #42, opened 2017-07-06
Make Function.prototype executable.

NeilFraser:


server/interpreter.js
```diff
@@ -542,6 +546,7 @@ Interpreter.prototype.initObject = function(scope) {
  */
 Interpreter.prototype.initArray = function(scope) {
   var thisInterpreter = this;
+  this.ARRAY = this.createArray(this.OBJECT);
```
cpcallen: Maybe add comment on line above here as in initDate, initRegExp etc.

server/interpreter.js
```diff
@@ -289,6 +287,12 @@ Interpreter.prototype.initFunction = function(scope) {
   var FunctionConst = this.createNativeFunction(wrapper, this.FUNCTION);
   this.addVariableToScope(scope, 'Function', FunctionConst);
 
+  // Configure Function.prototype.
+  this.FUNCTION.nativeFunc = function() {};
+  this.FUNCTION.id = this.functionCounter_++;
+  this.setProperty(this.FUNCTION, 'length', 0,
+      Interpreter.READONLY_DESCRIPTOR);
```
cpcallen: Why not just do

```JS
  this.FUNCTION = this.createNativeFunction(function() {}, this.OBJECT);
```
when creating it in initGlobalScope?

NeilFraser: Confirmed.  That works.

server/interpreter.js
NeilFraser: Done.

PR #46, opened 2017-07-07
Create intrp.Object.prototype.addNativeMethod (and use it)

cpcallen:


cpcallen:
These changes were inadvertently included in PR #48, so this earlier PR is no longer needed.

PR #49, opened 2017-07-07
Bug hunt

NeilFraser:


server/interpreter.js
```diff
@@ -2571,8 +2571,8 @@ Interpreter.prototype['stepCallExpression'] = function() {
       return;
     } else if (func.eval) {
       var code = state.arguments_[0];
-      if (!code) {  // eval()
-        state.value = undefined;
+      if (typeof code !== 'string') {  // eval()
+        state.value = code;
```
cpcallen: It wasn't clear what this is doing, but then I realised that this is the `eval(new String('1+1')) === 'new String(1+1)'` case.  Maybe a comment?

PR #51, opened 2017-07-07
Fix setProperty when overwriting a previously defined property.

NeilFraser:


server/interpreter.js
```diff
@@ -1522,28 +1522,45 @@ Interpreter.prototype.setProperty = function(obj, name, value, opt_descriptor) {
                         "', object is not extensible");
   }
   if (opt_descriptor) {
+    var previouslyDefined = name in obj.properties;
     // Define the property.
     obj.properties[name] = value;
-    if (!opt_descriptor.configurable) {
+    if (previouslyDefined ? opt_descriptor.configurable === false :
+        opt_descriptor.configurable !== true) {
```
cpcallen: Should be testing truthiness here (or make sure that any caller is only passing in `boolean`/`undefined`).

PR #56, opened 2017-07-10
Go faster

cpcallen:
About 8% faster.


cpcallen:
Verbal SGTM from @NeilFraser

PR #57, opened 2017-07-10
Monday

NeilFraser:


server/codecity.js
```diff
@@ -0,0 +1,122 @@
...
+  } catch (e) {
+    console.log('Syntax error in parsing JSON: ' + filename);
+    process.exit(1);
+  }
+
+  // Load the interpreter.
+  interpreter = new Interpreter();
+  Serializer.deserialize(contents, interpreter);
+  console.log('Database loaded: ' + filename);
+  // TODO: Run the interpreter.
```
cpcallen: Surely this should happen after setting up the checkpointing timeout?

server/codecity.js
```diff
@@ -0,0 +1,122 @@
...
+  var json = Serializer.serialize(interpreter);
+  // JSON.stringify(json) would work, but adding linebreaks so that every
+  // object is on its own line makes the output more readable.
+  var text = [];
+  for (var i = 0; i < json.length; i++) {
+    text.push(JSON.stringify(json[i]));
+  }
+  text = '[' + text.join(',\n') + ']';
+
+  var filename = (new Date()).toISOString().replace(/:/g, '-') + '.city';
```
cpcallen: +1 for UTC!

server/serialize.js
```diff
@@ -295,14 +299,14 @@ function objectHunt_(node, objectList) {
...
+};
 
-if (typeof module !== 'undefined') { // Node.js
-  Interpreter = require('./interpreter.js');
-  exports.deserialize = deserialize;
-  exports.serialize = serialize;
+if (typeof module !== 'undefined') {  // Node.js
+  Interpreter = require('./');
+  exports.deserialize = Serializer.deserialize;
+  exports.serialize = Serializer.serialize;
```
cpcallen: Why not just write `exports = Serializer`?

server/serialize.js
```diff
@@ -295,14 +299,14 @@ function objectHunt_(node, objectList) {
...
   }
-}
+};
 
-if (typeof module !== 'undefined') { // Node.js
-  Interpreter = require('./interpreter.js');
-  exports.deserialize = deserialize;
-  exports.serialize = serialize;
+if (typeof module !== 'undefined') {  // Node.js
+  Interpreter = require('./');
```
cpcallen: Now that there is actually a server 'main', we should probably change the entry in package.json to point to it, and continue to import `interpreter.js` explicitly.

server/autoexec.js
```diff
@@ -224,20 +224,43 @@ Object.defineProperty(Array.prototype, 'some',
...
+        if (comp > 0) {
+          r.push(this[i]);
+        } else if (comp < 0) {
+          l.push(this[i]);
+        } else {
+          e.push(this[i]);
         }
       }
-      if (changes <= 1) break;
+      return l.sort(opt_comp).concat(e, r.sort(opt_comp));
```
NeilFraser: Ugh, good point.  I should have caught that.  Reverting the sort (it's not important to current requirements, I only added it because it was a quick copy-paste).

server/codecity.js
```diff
@@ -0,0 +1,122 @@
...
+  } catch (e) {
+    console.log('Database not found.\nUsage: node ' + process.argv[1] +
+                ' <DB directory>');
+    process.exit(1);
+  }
+
+  // Find the most recent database file.
+  files.sort();
+  for (var i = files.length - 1; i >= 0; i--) {
+    if (files[i].match(/^\d\d\d\d-\d\d-\d\dT\d\d-\d\d-\d\d(\.\d)?\d?\d?Z?\.city$/)) {
```
NeilFraser: Added {4} and {1-3} to the regex, though {2} is longer than what it replaces.
Don't have any strong opinions regarding the time separator (other than disliking :, / or nothing).

server/codecity.js
NeilFraser: Comment moved.

server/serialize.js
NeilFraser: Because I have no idea what I'm doing?

server/serialize.js
NeilFraser: Done.

cpcallen:
Subject to verifying that the test all still pass: LGTM.

PR #58, opened 2017-07-13
Improve serialization

cpcallen:
More tests (including a benchmark of post-roundtrip performance); fewer bugs.

### DIFF
--- a/server/tests/run.js
+++ b/server/tests/run.js
@@ -24,6 +24,8 @@
  */
 'use strict';
 
+const fs = require('fs');
+
 /**
  * Class that records benchmark results; much like Go's testing.B type.
  * @constructor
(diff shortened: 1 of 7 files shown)


server/tests/run.js
```diff
@@ -175,5 +180,16 @@ function runTests() {
 
 /********************************************************************/
 
-runTests();
-runBenchmarks();
+/**
+ * Get list of test filenames matching a particular regexp.
+ * @param {!RegExp} pattern RegExp to match.
+ * @return {Array} List of filenames.
```
NeilFraser: {!Array<string>} I think.

server/tests/serialize_bench.js
```diff
@@ -0,0 +1,84 @@
...
+        b = sum;
+      }
+    }
+    for(var i = 0; i < 10000; i++) {
+      var result = [];
+      fibonacci(78, result);
+    }
+    result;
+  `;
+  runInterpreterBench(b, name, src);
```
NeilFraser: b is a B or null, according to the JSDoc.  But runInterpreterBench only accepts a B according to its JSDoc.   Either add a `if (b)` check around this call, or require b be non-null in this function's JSDoc.

server/tests/serialize_test.js
```diff
@@ -24,62 +24,124 @@
...
+ * - Run new instance to completion.
+ * - Append src2 (if specified) and run that to completion.
+ * - Verify result is as expected.
+ * @param {!T} t The test runner object.
+ * @param {string} name The name of the test.
+ * @param {string} src1 The code to be evaled before serialization.
+ * @param {string} src2 The code to be evaled after serialization.
+ * @param {number|string|boolean|null|undefined} expected The expected
+ *     completion value.
+ * @param {number=} stesp How many steps to run before serializing
```
NeilFraser: stesp?

cpcallen:
All fixed, I believe…

PR #61, opened 2017-07-17
Remove getScope

cpcallen:
Instead, add scope parameter to {get,set}Value and
{get,set}ValueFromScope.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1575,7 +1575,7 @@ Interpreter.prototype.deleteProperty = function(obj, name) {
 
 /**
  * Retrieves a value from the scope chain.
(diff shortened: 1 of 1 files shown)


server/interpreter.js
```diff
@@ -1574,24 +1574,12 @@ Interpreter.prototype.deleteProperty = function(obj, name) {
...
-  var scope = this.stateStack[this.stateStack.length - 1].scope;
-  if (!scope) {
-    throw Error('No scope found.');
-  }
-  return scope;
-};
-
-/**
  * Retrieves a value from the scope chain.
+ * @param {!Interpreter.Scope} scope Scope to write to.
```
NeilFraser: ...read from.

PR #62, opened 2017-07-17
Introduce threads

cpcallen:
Introduce multiple threads and a (very) simplistic scheduler.

This implementation imposes a 10-15% performance penalty (mostly due to the extra cost of resolving this.thread.stateStack instead of this.stateStack in pushNode_) but a fix is in hand and will follow in a subsequent PR.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1013,7 +1013,7 @@ Interpreter.prototype.initDate = function(scope) {
     })(functions[i]);
     this.DATE.addNativeMethod(functions[i], wrapper);
   }
-  functions = ['toLocaleDateString', 'toLocaleString',
+  var functions = ['toLocaleDateString', 'toLocaleString',
                'toLocaleTimeString'];
   for (var i = 0; i < functions.length; i++) {
     wrapper = (function(nativeFunc) {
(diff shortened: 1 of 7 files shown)


server/interpreter.js
```diff
@@ -123,46 +119,50 @@ Interpreter.STEP_ERROR = {};
...
-  if (!code || code['type'] !== 'Program') {
-    throw Error('Expecting new AST to start with a Program node.');
-  }
-  this.populateScope_(code, state.scope);
-  // Append the new program to the old one.
-  for (var i = 0, node; (node = code['body'][i]); i++) {
-    state.node['body'].push(node);
+Interpreter.prototype.createThread = function(code) {
+  var id = this.threads.length;
+  var thread = new this.Thread(id, code);
```
NeilFraser: Normalize code to AST here, rather than later?

server/interpreter.js
```diff
@@ -123,46 +119,50 @@ Interpreter.STEP_ERROR = {};
...
+  this.threads.push(thread);
+  this.schedule();
```
NeilFraser: At least add a TODO indicating that this needs to be moved.

server/interpreter.js
```diff
@@ -1837,6 +1835,82 @@ Interpreter.prototype.pushNode_ = function(node) {
 //   small performance improvement in V8.
 
 /**
+ * @constructor
+ * @param {number} id
+ * @param {string|!Object} code
```
NeilFraser: Can haz more words?

server/interpreter.js
cpcallen: Yes.  Done.

server/interpreter.js

server/interpreter.js
cpcallen: I now have some JSDoc output to look at, but it makes me think we should delay worrying about this for now.

PR #63, opened 2017-07-17
Fix delete 'hello' and other bugs.

NeilFraser:


server/interpreter.js
```diff
@@ -3097,18 +3080,21 @@ Interpreter.prototype['stepUnaryExpression'] = function(stack, state, node) {
   } else if (node['operator'] === '~') {
     value = ~value;
   } else if (node['operator'] === 'delete') {
+    // If value is not an array, then it is a primitive.  If so, return true.
```
cpcallen: This comment looks misplaced, but perhaps is actually mis-worded.  How about: "It's only possible to delete expressions which evaluate to a reference".

server/interpreter.js
```diff
@@ -3097,18 +3080,21 @@ Interpreter.prototype['stepUnaryExpression'] = function(stack, state, node) {
...
-      var name = value[1];
-    } else {
-      var obj = state.scope;
-      var name = value;
-    }
-    value = this.deleteProperty(obj, name);
-    if (!value) {
-      this.throwException(this.TYPE_ERROR, "Cannot delete property '" +
-                          name + "' of '" + obj + "'");
+      // If the obj isn't an object, then it is a scope.  If so, return true.
```
cpcallen: Can it ever actually be a scope?  Won't Acorn catch that?

PR #67, opened 2017-07-18
Green connect, reload icon tooltip, save tab in cookie.

NeilFraser:


www-static/client/connect.js
```diff
@@ -194,17 +194,22 @@ CCC.init = function() {
   document.body.addEventListener('click', function() {
     CCC.postToAllFrames({'mode': 'blur'});
   }, true);
-  CCC.tab('world');
+  CCC.tab();
   CCC.schedulePing(0);
   // Firefox sometimes caches the disabled value on reload.
```
cpcallen: > `// Firefox sometimes caches the disabled value on reload.`

WTF?

www-static/client/connect.js
```diff
@@ -194,17 +194,22 @@ CCC.init = function() {
...
 
 /**
  * Switch between world and log views.
- * @param {string} mode Either 'world' or 'log'.
+ * @param {string=} mode Either 'world' or 'log', or undefined.
  */
 CCC.tab = function(mode) {
+  if (!mode) {
+    // Check for a cookie preference.
+    var m = document.cookie.match(/(?:^|;\s*)tab=(\w+)(?:;|$)/);
```
cpcallen: Ugh—is that really how one is supposed to find a cookie value from JS?  How wretched. 
 On the other hand, cute regexp: `?:^|`.

www-static/client/connect.js
```diff
@@ -194,17 +194,22 @@ CCC.init = function() {
...
+    mode = m ? m[1] : 'world';
```
cpcallen: Is it worth validating the mode?

www-static/client/world.js
```diff
@@ -410,7 +410,8 @@ CCC.World.prerenderPanorama = function(node) {
 CCC.World.connectPanel = function(node) {
   var df = document.createDocumentFragment();
   var div = document.createElement('div');
-  div.className = 'systemDiv';
+  div.className =
+      (CCC.Common.isConnected ? 'connectDiv' : 'disconnectDiv');
   var msg = CCC.World.getMsg(node.isConnected ?
```
cpcallen: Why is the line above conditional on `ccc.Common.isConnected` but this one on `node.isConnected`?  That smells bad (or at least like it deserves a comment).

PR #68, opened 2017-07-18
Use proper constructor to create fake AST node for EvalExpression

cpcallen:
Nothing was broken previously, but this should avoid any problems in
future.

Also fixed type decls for Closure Compiler (so that it would have
caught this earlier).


cpcallen:
Going to hold off on this for the moment.

PR #70, opened 2017-07-20
Support string.replace with function argument.

NeilFraser:


server/autoexec.js
```diff
@@ -253,6 +250,40 @@ Object.defineProperty(Array.prototype, 'toLocaleString',
   }
 });
 
+// Add a polyfill to handle replace's second argument being a function.
+(function() {
+  var replace_ = String.prototype.replace;
```
cpcallen: At such point as we have the `new` hack working we should use that here instead of a closure.  It will make it easier to understand*/edit the polyfill.

(*) at least if you understand the `new` hack, anyway!

server/autoexec.js
```diff
@@ -253,6 +250,40 @@ Object.defineProperty(Array.prototype, 'toLocaleString',
...
+        subs.push([m.index, m[0].length, inject]);
+        m = substr.global ? substr.exec(str) : null;
+      }
+      for (var i = subs.length - 1; i >= 0; i--) {
+        str = str.substring(0, subs[i][0]) + subs[i][2] +
+            str.substring(subs[i][0] + subs[i][1]);
+      }
+    } else {                         // string.replace(string, function)
+      var i = str.indexOf(substr);
+      if (i !== -1) {
```
cpcallen: Shouldn't there be a loop here?  (But not a naïve one: output from `newSubstr()` shouldn't be rescanned...)

server/interpreter.js
```diff
@@ -933,7 +934,7 @@ Interpreter.prototype.initString = function(scope) {
   this.STRING.addNativeMethod('split', wrapper);
 
   wrapper = function(regexp) {
-    regexp = regexp ? regexp.data : undefined;
+    regexp = regexp ? regexp.regexp : undefined;
```
cpcallen: Oops.  Good catch.

server/interpreter.js
```diff
@@ -943,14 +944,14 @@ Interpreter.prototype.initString = function(scope) {
...
     return this.search(regexp);
   };
   this.STRING.addNativeMethod('search', wrapper);
 
-  wrapper = function(substr, newSubStr) {
-    // TODO: Rewrite as a polyfill to support function replacements.
-    return this.replace(substr, newSubStr);
+  wrapper = function(substr, newSubstr) {
+    // Support for function replacements is the responsibility of a polyfill.
+    return String(this).replace(substr.regexp || substr, newSubstr);
```
cpcallen: Please use `substr instanceof intrp.RegExp` check here.

server/autoexec.js
NeilFraser: Only one replacement is made:
`  'aaa'.replace('a', function(){return 'x';}) -> 'xaa'`
A global regexp is the only way to do multiple replacements.

cpcallen: Huh.  I read the ES5 spec several times and it's still not apparent that you are correct.  Fortunately the MDN spec is reasonably clear about this: "`substr` (pattern):
A String that is to be replaced by newSubStr…  Only the first occurrence will be replaced".

server/interpreter.js
NeilFraser: I think you mean thisInterpreter.RegExp.

cpcallen: Only until I get round to s/thisInterpreter/intrp/g as we previously discussed...  :-)

server/tests/interpreter_test.js
```diff
@@ -710,6 +710,28 @@ exports.testLegalArrayIndexLength = function(t) {
...
+ * @param {!T} t The test runner object.
+ */
+exports.testStringReplace = function(t) {
+  var cases = [
+    // replace(string, string)
+    ["'xxxx'.replace('xx', 'y')", "yxx"],
+    // replace(regexp, string)
+    ["'xxxx'.replace(/(X)\\1/ig, 'y')", "yy"],
+    // replace(string, function)
+    ["'xxxx'.replace('xx', function(a,b,c,d,e){return ['[',a,',',b,',',c,',',d,',',e,']'].join('')})", "[xx,0,xxxx,,]xx"],
```
cpcallen: Why not `return arguments.join();`?

server/tests/interpreter_test.js
```diff
@@ -710,6 +710,28 @@ exports.testLegalArrayIndexLength = function(t) {
...
+    // replace(regexp, function)
+    ["'xxxx'.replace(/(X)\\1/ig, function(a,b,c,d,e){return ['[',a,',',b,',',c,',',d,',',e,']'].join('')})", "[xx,x,0,xxxx,][xx,x,2,xxxx,]"],
```
cpcallen: (and likewise)

PR #71, opened 2017-07-21
Add command parser.

NeilFraser:


server/command.js
```diff
@@ -0,0 +1,243 @@
...
+    dobj: dobj,
+    prepstr: prepstr,
+    iobjstr: iobjstr,
+    iobj: iobj
+  };
+};
+
+$.utils.command.execute = function(command, user) {
+  // Parse and execute a user's command.
+  // Return true if successful, false if not.
```
cpcallen: TODO for security checking / permissions setting?

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  var spec = $.utils.command.parse(command, user);
```
cpcallen: I'd avoid `spec` here for reasons which will become clear in a moment.  How about `command`?

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+    if (!host) {
+      continue;
+    }
+    // Check every verb on each object for a match.
+    var func, verb, dobj, prepstr, iobj;
+    for (var prop in host) {
+      func = host[prop];
+      if (typeof func === 'function' && (prepstr = func.prepstr) &&
+          (dobj = func.dobj) && (iobj = func.iobj && (verb = func.verb))) {
+        // This is a verb.
```
cpcallen: * How about `prepspec` instead of `prepstr` (etc.)?
* What if `verb` were the only required one to flag verbiness, and the other default to 'any' if absent?
* Bonus feature: if `dobjspec` or `iobjspec` is not a string but a $.physical object then *that particular object is the only valid one for this verb* (or: only children of it are valid).
* So much avoidable nesting.  How about:

```JS
      if (typeof func !== 'function' || !(verb = func.verb)) {
        continue;
      }
      dobj = func.dobjspec;
      /* etc. */
      if (prepspec !== 'any' || prepspec !== spec.prepstr ||
          prepspec === 'none' && spec.prepstr ||
          dobj !== 'any' || /* etc. */ ) {
        continue;
      }
      host[prop](spec);
      return true;
```
* Probably better to do the `spec.verb.search(verb) === 0` check first, since that is the least likely to match.
* Do you mean to use `==` instead of `===`?

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  return false;
+};
+
+$.utils.match = function(str, context) {
+  // Attempt to match a string with the environmental context.
+  // Context can be anything (including null) but is normally a place or a user.
+  // Returns an array of all matching objects.
+  str = str.trim();
+  // First, check for matches against universally accessible things.
+  var m = $(str);
```
cpcallen: Yes, if we're doing `$()` here then `$()` had better not eval.  (We want to be able to have non-programmers, after all!)

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  return [];
+};
+
+$.utils.match.strength = function(str, obj) {
+  // On a score of zero to three, how well does str match obj?
+  // 0: No match.
+  // 1: Partial name or alias.
+  // 2: Perfect alias match.
+  // 3: Perfect name match.
+  if (!str || !obj) {
```
cpcallen: `if (typeof str !== 'string' || !$.physical.isPrototypeOf(obj)) {`

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+      if (contents[i] !== context) {
+        matchObj(contents[i]);
+      }
+    }
+  }
+  // Return the highest level bin.
+  if (nameMatches.length) {
+    return nameMatches;
+  }
+  if (aliasMatches.length) {
```
cpcallen: It's sort of unfortunate that zero-length arrays are truthy, or this could be written so much more concisely.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  }
+  var location = context.location;
+  if (context instanceof $.user && location instanceof $.physical) {
+    // Try the user's place next.
+    matchObj(location);
+    // Check the user's place's contents.
+    var contents = context.contents();
+    for (var i = 0; i < contents.length; i++) {
+      if (contents[i] !== context) {
+        matchObj(contents[i]);
```
cpcallen: Why not quickly concatenate together a list of {context, @context.contents, context.location, @context.location.context} (to use MOO syntax) and then loop over that?

server/command.js
```diff
@@ -0,0 +1,243 @@
+var $ = function(str) {
+  // Parse the string to extract a reference.
+  // TODO: Support notations starting with @roomname and ~username.
```
NeilFraser: Good point.  Something we'll have to keep an eye on.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  var root = m[1];
+  var suffix = m[2];
+  if (root === '$') {
+    root = $;
+  } else if (root === '~' || root === 'me') {
+    root = user;
+  } else if (root === '@' || root === 'here') {
+    root = user.location;
+  } else {
+    throw Error("Can't happen.  Regex is too liberal.");
```
NeilFraser: Correct.  Can't happen.  But I want to know about it if it does.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  }
+  if (!suffix) {
+    return root;
+  }
+  // Clear no-longer needed local variables so they don't leak into the eval.
+  str = undefined;
+  m = undefined;
```
NeilFraser: eval contents now restricted so that variable matching is not possible.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  return eval('(root)' + suffix);
```
NeilFraser: Done.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+};
+
+
+$.utils = {};
+$.utils.command = {};
```
NeilFraser: A flattened $ hierarchy would have a thousand names at the root level and be non-explorable.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+
+$.utils.command.prepositions = Object.create(null);
+$.utils.command.prepositions['with'] = 'with/using';
+$.utils.command.prepositions['using'] = 'with/using';
+$.utils.command.prepositions['at'] = 'at/to';
```
NeilFraser: > Why not initialise $.utils.command.prepositions and .prepositionsRegExp from an array literal?

This is code that will be thrown away once a minimal database is self-sustaining.  Such an initialisation function would be lost.

> Or omit the former, since it does not appear to be used.

Oops, good point, they are absolutely needed in (and were written for) $.utils.command.execute.  Added.

> And if we think it's likely that the list of prepositions will need to be modified in future 

I don't think we ever modified the preps in Moo Canada.  With the exception of when we attempted to translate everything to French.  But I'm not willing to write the entire database in an easily localizable form since that would mean harder to read code, slower execution, and vastly more work.  Times ten if one wants to think about Arabic and Hebrew.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+
+$.utils.command.prepositionsRegExp = new RegExp(
+    '(?:^|\\s)(with|using|upon|underneath|under|to|through|over|out +of|' +
+    'onto|on +top +of|on|off +of|off|is|into|inside|in +front +of|' +
+    'in|from +inside|from|for|beside|beneath|behind|at|as|about)(?:\\s|$)');
+
+$.utils.command.parse = function(command, user) {
+  // Parse a user's command into components:
+  // verb, argstr, args, dobjstr, dobj, prepstr, iobjstr, iobj
+  var args = command.split(/ +/);
```
NeilFraser: Done.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+  var dobj = null;
+  var iobjstr = '';
+  var iobj = null;
+  var prepstr = '';
+  var m = argstr.match($.utils.command.prepositionsRegExp);
+  if (m) {
+    prepstr = m[1];
+    var i = argstr.indexOf(prepstr);
+    if (i === -1) {
+      throw RangeError('Impossible preposition: ' + prepstr);
```
NeilFraser: Shouldn't be possible.  But if there's a bug in the regex I'd rather know about it than do bizarre things with i === -1.

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+    dobjstr = argstr.substring(0, i - 1);
+    iobjstr = argstr.substring(i + prepstr.length + 1);
+    prepstr = prepstr.replace(/ +/g, ' ');
```
NeilFraser: This would make it more difficult to support a feature Lambda Moo has which is quoted objects: `say "I'm at home" to Max`
In this case we want the prep to be "to" not "at".

server/command.js
```diff
@@ -0,0 +1,243 @@
...
+    dobjstr = argstr;
+  }
+  if (dobjstr) {
+    var m = $.utils.match(dobjstr, user);
+    if (!m.length) {
+      dobj = undefined;
+    } else if (m.length === 1) {
+      dobj = dobj[0];
+    } else {
+      dobj = ReferenceError(dobj.length + ' matches.');
```
NeilFraser: Good idea.  Done.

server/command.js
NeilFraser: Done.

server/command.js
NeilFraser: Agreed that spec is a horrible name.  But overwriting command would mean losing the original string.  Currently we don't use it, but clobbering it precludes the ability to create error message such as:
`I don't understand "loook north".`

server/command.js
NeilFraser: > How about prepspec instead of prepstr (etc.)?

Str actually means something, a string, whereas spec is meaningless.  My use of spec as a local variable here was a signal that I'd given up trying to create a meaningful name for that data object.  But at least it's local.  Let's not export that to users.

>What if verb were the only required one to flag verbiness, and the other default to 'any' if absent?

These properties will be created/populated automatically by the editor, so being sparse doesn't save the users any work.  And being sparse would mean that all functions that look at verbs would need to be more complicated.  I'm only seeing negatives here.

> Bonus feature: if dobjspec or iobjspec is not a string but a $.physical object then that particular object is the only valid one for this verb (or: only children of it are valid).

Writing the web-based editor where that's readable and writable would be a challenge.  I'm totally down for having d/iobjstr be a standard '$.foo.bar' string though.  But that's something we can add later.

> So much avoidable nesting.Probably better to do the spec.verb.search(verb) === 0 check first, since that is the least likely to match.

Agreed.

> Do you mean to use == instead of ===?

Good catch.

server/command.js
NeilFraser: $() has been downgraded so that it can only do property lookups.  We can look at restoring function calls at a later time.

server/command.js
NeilFraser: That would be an O(n^2) solution.  This is O(n).

server/command.js
NeilFraser: Visibly unfortunate in this case, invisibly fortunate in other cases.

server/command.js
NeilFraser: This was deliberate.  For str, any false values (like undefined and null) should not match.  For obj, there's no reason why a non-physical object with the appropriate interfaces shouldn't be acceptable (like how array functions work on non-arrays).

core/command.js
```diff
@@ -19,10 +19,14 @@ var $ = function(str) {
...
     return root;
   }
-  // Clear no-longer needed local variables so they don't leak into the eval.
-  str = undefined;
-  m = undefined;
-  return eval('(root)' + suffix);
+  // This regex has two main groups:
+  // 1) match .foo
+  // 2) match [42] or ['bar'] or ["baz"]
+  if (suffix.search(/^((\s*\.\s*[A-Za-z$_][A-Za-z0-9$_]*\s*)|(\s*\[\s*(\d+|\'([^\'\\]*(\\.[^\'\\]*)*)\'|"([^"\\]*(\\.[^"\\]*)*)")\s*\]\s*))+$/) === 0) {
```
cpcallen: Strictly speaking this is too strict: where you have `[A-Za-z$_]` for identifier names, the spec allows *any Unicode letter*.  But AFAICT JS does not support character classes like `[[:letter:]]`, so doing better than this is not really very straight forward.

server/command.js
cpcallen: > Str actually means something, a string, whereas spec is meaningless.

Ahh, I have not been clear.  Here is what I was thinking: when matching the input "kick max" against the vert ['kick', 'this', 'none', 'none']:

- 'this' is the dobjspec
- 'max' is the dobjstr
- #833 is the dobj

i.e., use 'spec' to clarify that we are talking about one of the set 'this', 'none' or 'any', rather than the string that gets matched against it.

> These properties will be created/populated automatically by the editor, so being sparse doesn't save the users any work.

Fair point.  (I was also thinking about avoiding traceback in the case a function is created by code (or by eval) with a partial verb spec, but reading the code again I see that won't be a problem.)

>> So much avoidable nesting…
> Agreed

And yet I still see much nesting.  The point was that all the `if` statements can be combined into a single one with `continue;` in the body.

server/command.js
cpcallen: Don't understand this.  How is it harder to support quoted strings if using a regexp to break argstr up?

server/command.js
cpcallen: Yeah: I did not mean to suggest overwriting command.  I was thinking more of what the canonical name of the parameter for verb functions should be, rather than of the local variable in this function.

server/command.js
cpcallen: No; it should be O(m * n), where m is the number of source lists (fixed at 4), and n is total number of items in the concatenated list.

server/command.js
cpcallen: Dunno; in Perl empty lists are falsy and that doesn't seem to cause any trouble.  (They aren't objects, though.)

server/command.js
cpcallen: No problem rejecting empty strings, but as-is this code will crash if `str` is a number, or if `obj` has a `null` prototype (or indeed fails to have `.aliases` defined as an array).

core/command.js
cpcallen: Oh, also: there are a few spots where you have written `\'`, where I think you just mean `'`.

PR #74, opened 2017-07-27
Runtime changes in preparation for networking

cpcallen:
Various changes that are necessary to support networking or just simply continuous operation of the interpreter:

- Run tests asynchronously
- Have start and stop methods on interpreter instances (that do what they say on the tin)
- Ability to specify ‘this’ value when creating a new thread to execute a function call (i.e., can put *method* calls in own thread).

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -184,7 +184,7 @@ Interpreter.prototype.createThread = function(runnable) {
  * none.  If there are additionally no BLOCKED threads left (i.e.,
  * there are no non-ZOMBIE theads at all) it will also set .done to
  * true.
- * @return {number} See above.
+ * @return {number} See description.
  */
 Interpreter.prototype.schedule = function() {
   var now = this.now();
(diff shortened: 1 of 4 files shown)


server/interpreter.js
```diff
@@ -290,7 +333,38 @@ Interpreter.prototype.run = function() {
...
   }
-  return !this.done;
+  if (t === Number.MAX_VALUE) {
+    return this.done ? 0 : -1;
+  }
+  return t;
+};
+
+/**
+ * Use setTimout to repeatedly call .run() until there are no more
```
NeilFraser: setTimEout

server/tests/interpreter_test.js
```diff
@@ -781,14 +778,70 @@ exports.testThreading = function(t) {
...
+  function snooze(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+  var intrp = new Interpreter;
+  var name = 'testStartStop';
+  var src = `
+      var x = 0;
+      while (true) {
+          suspend(10);
+          x++;
```
NeilFraser: Go style indenting?

server/interpreter.js
```diff
@@ -171,23 +174,55 @@ Interpreter.prototype.createThread = function(runnable) {
...
- * @return {boolean} True iff there is a runnable thread.
+ * Create a new thread to execute a particular function call.
+ * @param {!Interpreter.prototype.Function} fun Function to call.
+ * @param {Interpreter.Value} funcThis value of 'this' in function call.
+ * @param {!Array<Interpreter.Value>} args Arguments to pass.
+ * @param {number=} runAt Time at which thread should begin execution
+ *     (defaults to now).
+ * @return {number} thread ID.
+ */
+Interpreter.prototype.createThreadForFuncall =
```
NeilFraser: Funcall == FunCall or FuncAll?  Given that I think you mean the former, and functions are elsewhere referred to as func, suggest renaming to FuncCall.  Less fun though.

cpcallen: In the world of functional programming the spelling is invariably "funcall" (no double C), after the lisp keyword of the same name.  But FuncCall will probably be clearer here.

server/interpreter.js

server/tests/interpreter_test.js
cpcallen: Done.

PR #78, opened 2017-08-09
Provide source code in output of Function.prototype.toString

NeilFraser:


server/interpreter.js
```diff
@@ -2455,6 +2455,37 @@ Interpreter.prototype.installTypes = function() {
...
+        var node = state.node;
+        if (node['type'] !== 'CallExpression' && stack.length) {
+          continue;
+        }
+        var code = intrp.getSource(i);
+        var lineStart = code.lastIndexOf('\n', node['start']);
+        if (lineStart === -1) {
+          lineStart = 0;
+        } else {
+          lineStart++;
```
cpcallen: The `if` is redundant.

PR #81, opened 2017-08-14
Running server

NeilFraser:


server/tests/db/test_01_es5.js
```diff
@@ -0,0 +1,1739 @@
...
+};
+
+tests.demethodedCall = function() {
+  var o = { f: function() { return this; }};
+  var g = o.f;
+  console.assert(g() === undefined, 'demethodedCall');
+};
+
+tests.testThis = function() {
+  console.assert(this === tests, 'testThis');
```
cpcallen: This is not the same as the original test, which tests the value of `this` at the top level (i.e., that it is `undefined` rather an alias of `window` or the like).

server/tests/db/test_01_es5.js
```diff
@@ -0,0 +1,1739 @@
...
+  console.assert(a[0] === 'hi' && a.length === 1,
+                 'arrayLengthWithNonConfigurableProps');
+};
+
+tests.compValEmptyBlock = function() {
+  console.assert(eval('{}') === undefined, 'compValEmptyBlock');
+};
+
+tests.undefined = function() {
+  console.assert(undefined === undefined, 'undefined');
```
cpcallen: I'm not really sure this test is testing anything useful, whereas the original was.  Maybe use `void` to obtain a canonical `undefined` value to compare with the value of the `undefined` variable?

server/tests/db/test_01_es5.js
```diff
@@ -0,0 +1,1739 @@
...
+};
+
+tests.unaryVoid = function() {
+  var x = 70;
+  console.assert((undefined === void x++) && x === 71, 'unaryVoid');
```
cpcallen: Oh look...

server/tests/db/test_01_es5.js
```diff
@@ -0,0 +1,1739 @@
...
+};
+
+tests.unaryPlusMinus = function() {
+  console.assert(+'72' === 72, 'unaryPlus');
+  console.assert(-73 === -73, 'unaryPlus');
```
cpcallen: There are at least two issues with this line.

server/tests/db/test_01_es5.js
```diff
@@ -0,0 +1,1739 @@
...
+  } catch (e) {
+    console.assert(e.name === 'TypeError', 'RegExpPrototypeTestApplyNonRegExpThrowsError');
+  }
+};
+
+//////////////////////////////////////////////////////////////
+// Other tests
+
+tests.newHack = function() {
+  console.assert((new 'Array.prototype.push') === Array.prototype.push, 'xxx');
```
cpcallen: More adult material?

server/tests/db/test_20_reboot.js
```diff
@@ -0,0 +1,6 @@
+
+$.system.shutdown();
+
+$.system.log('Benchmarking resurrected fibonacci10k...');
+test_fibonacci10k();
+setTimeout($.system.shutdown, 1000);
```
cpcallen: Why setTimeout here?

PR #83, opened 2017-08-15
Create minimal database.

NeilFraser:


minimal/core_01_minimal.js
```diff
@@ -0,0 +1,258 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Minimal database for Code City.
+ * @author fraser@google.com (Neil Fraser)
+ */
+
+var user = null;
```
cpcallen: This may work well enough for now, but note that both setTimeout and network activity are going to be incompatible with this approach.

minimal/core_01_minimal.js
```diff
@@ -0,0 +1,258 @@
...
+$.system.log = new '$.system.log';
+$.system.checkpoint = new '$.system.checkpoint';
+$.system.shutdown = new '$.system.shutdown';
+$.system.connectionWrite = new 'connectionWrite';
+$.system.connectionClose = new 'connectionClose';
+
+
+// Prototype physical object: $.physical
+$.physical = {};
+$.physical.name = 'Prototype physical object';
```
cpcallen: Naming convention: "Array prototype" or "Prototype array"?  I think the former is more usual, so hence this would be "Physical object prototype" (which also puts the more distinguishing part of the name at the beginning).

minimal/core_01_minimal.js
```diff
@@ -0,0 +1,258 @@
...
+    }
+    user.tell('Contents: ' + text.join(', '));
+  }
+};
+$.physical.look.dobj = 'this';
+
+
+// Prototype thing: $.thing
+$.thing = Object.create($.physical);
+$.thing.name = 'Prototype thing';
```
cpcallen: See above.

minimal/core_01_minimal.js
```diff
@@ -0,0 +1,258 @@
...
+
+$.connection = {};
+$.connection.onConnect = function() {
+  this.user = null;
+  this.write('Welcome.  Type name of user to connect as (Alpha or Beta).');
+};
+$.connection.onReceive = function(text) {
+  if (this.user) {
+    user = this.user;
+    $.execute(text);
```
cpcallen: There is no guarantee that calls to onRecieve will correspond with '\n's.  Socket read boundaries are arbitrary.

minimal/core_01_minimal.js
```diff
@@ -0,0 +1,258 @@
...
+    return;
+  }
+  text = text.trim().toLowerCase();
```
cpcallen: Comment here that remainder of function just handles login?

minimal/core_01_minimal.js
```diff
@@ -0,0 +1,258 @@
...
+
+  var beta = Object.create($.user);
+  beta.name = 'Beta';
+  $.playerDatabase[beta.name.toLowerCase()] = beta;
+  beta.description = 'Mostly harmless.';
+  beta.moveTo(hangout);
+
+  var rock = Object.create($.thing);
+  rock.name = 'Rock';
+  rock.description = 'Cube shaped, made of granite.';
```
cpcallen: *Suspiciously* cube shaped.

PR #86, opened 2017-08-21
Fix try/catch/finally vs. break / continue / return / throw

cpcallen:
We unify the handling of break, continue, return and throw into a single function, .unwind(), which unwinds the stack back either to the target of the break or continue, or to the enclosing CallExpression (for return), but stops if a TryStatement is found.  A completion value (as described in §8.9 of the ES5.1 spec) is stored in state.cv.

If the abrupt completion is still pending when stepTryStatement has finished evaluating its subtrees it is dealt with by re-invoking .unwind() to continue propagating it up the stack.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -2084,6 +2084,18 @@ Interpreter.prototype.setValue = function(scope, ref, value) {
 };
 
 /**
+ * Completion Value Types.
+ * @enum {number}
+ */
+Interpreter.Completion = {
+  NORMAL: 0,
+  BREAK: 1,
+  CONTINUE: 2,
+  RETURN: 3,
+  THROW: 4,
+}
+
+/**
  * Throw an exception in the interpreter that can be handled by a
  * interpreter try/catch statement.  Can be called with either an
  * error class and a message, or with an actual object to be thrown.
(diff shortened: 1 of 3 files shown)


server/interpreter.js
```diff
@@ -2084,10 +2084,21 @@ Interpreter.prototype.setValue = function(scope, ref, value) {
...
+ * Completion Value Types.
+ * @enum {number}
+ */
+Interpreter.Completion = {
+  NORMAL: 0,
+  BREAK: 1,
+  CONTINUE: 2,
+  RETURN: 3,
+  THROW: 4,
+}
```
NeilFraser: Semicolon.  Unneeded comma on last line.

server/interpreter.js
```diff
@@ -2103,42 +2114,78 @@ Interpreter.prototype.throwException = function(value, opt_message) {
...
 /**
- * Throw an exception in the interpreter that can be handled by an
- * interpreter try/catch statement.  If unhandled, a real exception will
- * be thrown.
- * @param {Interpreter.Value} error Value being thrown.
+ * Unwind the stack to the innermost relevant enclosing TryStatement,
+ * For/ForIn/WhileStatement or Call/NewExpression.  If this results in
+ * the stack being completely unwound the thread will be terminated
+ * and an appropriate error being logged.
+ * @param {Interpreter.Completion} type Completion type.
```
NeilFraser: Should this be {!Interpreter.Completion}?

server/interpreter.js
```diff
@@ -2103,42 +2114,78 @@ Interpreter.prototype.throwException = function(value, opt_message) {
...
+    }
+    if (type === Interpreter.Completion.BREAK) {
+      if (label === undefined ?
+          (state.isLoop || state.isSwitch) :
+          (state.labels && state.labels.indexOf(label) !== -1)) {
+        stack.pop();
+        return;
+      }
+    } else if (type === Interpreter.Completion.CONTINUE) {
+      if (label == undefined ?
```
NeilFraser: Either `label === undefined` or `!label`.  The question is what should happen if label is empty string?

server/interpreter.js
```diff
@@ -2103,42 +2114,78 @@ Interpreter.prototype.throwException = function(value, opt_message) {
...
-    if (state.node['type'] === 'TryStatement') {
-      state.throwValue = error;
-      return;
+    switch (state.node['type']) {
+      case 'TryStatement':
+        state.cv = {type: type, value: value, label: label};
+        return;
+      case 'CallExpression':
+      case 'NewExpression':
+        switch (type) {
```
NeilFraser: More efficient:
if (type === Interpreter.Completion.RETURN) {
  state.value = value;
  return;
}
throw Error('Unsynatctic break/continue not rejected by acorn');

server/interpreter.js
cpcallen: Fixed.

server/interpreter.js
```diff
@@ -2847,25 +2894,11 @@ Interpreter.prototype['stepBlockStatement'] = function(stack, state, node) {
 };
 
 Interpreter.prototype['stepBreakStatement'] = function(stack, state, node) {
-  stack.pop();
-  var label = null;
+  var label = undefined;
   if (node['label']) {
     label = node['label']['name'];
   }
```
NeilFraser: var label = node['label'] && node['label']['name'];
or 
var label = node['label'] ? node['label']['name'] : undefined;

server/interpreter.js
cpcallen: I don't think so: Interpreter.Completion is an `@enum {number}`, so `null` should not be a valid argument and (just checking… Yes!) the compiler does correctly warn if you attempt to pass `null`.

server/interpreter.js
cpcallen: The spec defines labels as identifiers, so the empty string is not valid.  Label sets are either non-empty collections of labels, or the distinguished value `empty` (not an identifier spelled that way, but a special sentinel value).  It represents unlabelled breaks as "(break, empty, empty)"; I think we could represent them as either `''` (as I did in the Go interpreter) or `undefined` (as I do here) or even `null`.  I don't see any obvious reason to pick one over the others, other than that using `null` or `undefined` seems slightly clearer than just using the empty string.

As to this particular bit of *code*: it's probably best to use `!label` here as it will be most tolerant to incorrect input and is most clear about the particular information we are interested in (is there a label?)

server/interpreter.js
cpcallen: Not equivalent, as discussed…

server/interpreter.js
cpcallen: Fixed.

PR #88, opened 2017-08-22
Messages

NeilFraser:


demo/core_10_base.js
```diff
@@ -104,15 +104,11 @@ $.physical.moveTo = function(dest) {
...
     for (var i = 0; i < contents.length; i++) {
-      var user = contents[i];
-      if ($.user.isPrototypeOf(user)) {
-        var text = '<scene user="' + $.utils.htmlEscape(user.name) +
-            '" room="' + $.utils.htmlEscape(room.name) +
-            '" requested="false">\n' + scene + '</scene>';
-        user.tell(text);
+      var who = contents[i];
+      if ($.user.isPrototypeOf(who)) {
+        room.tellScene(who, false);
```
cpcallen: Naming inconsistency: `user.tell(msg)` tells to user, but `room.tellScene(who)` tells to who, not room.  Confusing.

Also: I would omit second parameter.

PR #89, opened 2017-08-23
Editor improvments

cpcallen:
Web server can now handle general query string parsing.

Edit command is changed to `edit <propertyKey> on <object>`.


cpcallen:
Verbal LGTM received from @NeilFraser.

PR #92, opened 2017-08-23
New Acorn + always strict

cpcallen:


cpcallen:
Verbal LGTM from @NeilFraser.

PR #100, opened 2017-08-30
JSON

NeilFraser:


www-static/index.html
```diff
@@ -2,15 +2,44 @@
...
-    <p>This directory contains static files which are served to the web using Nginx.</p>
+  <h1>
+    <svg xmlns="http://www.w3.org/2000/svg" width="95px" height="100px">
+      <path
+         d="m 1,90 0,-25 10,0 0,-8 9,0 0,-5 8,0 0,5 2,0 0,33 3,0 0,-13 2,0 0,-7 2,0 0,-32 1,0 0,-7 1,0 0,-6 2,0 0,-1 1,0 0,-12 1,0 0,-12 0,12 1,0 0,12 1,0 0,1 2,0 0,6 1,0 0,7 1,0 0,32 2,0 0,7 2,0 0,13 4,0 0,-32 5,0 0,-5 6,0 0,5 3,0 0,32 5,0 0,-42 3,0 0,-9 7,0 0,-2 5,0 0,2 2,0 0,51"
+         style="fill: none; stroke: #000; stroke-width: 1.5;" />
+    </svg>
+    Code City
+  </h1>
+  <p id="tagline">A community of young programmers.</p>
```
cpcallen: How about something less age-specific than "young"?  Perhaps "budding" or "aspiring"?

www-static/client/connect.css
```diff
@@ -68,6 +68,11 @@ tfoot {
   fill: #444;
 }
 
+.jfk-button {
+  border-top-left-radius: 0;
+  border-top-right-radius: 0;
```
cpcallen: Why only top corners?

demo/core_10_base.js
```diff
@@ -55,14 +55,40 @@ $.utils.commandMenu = function(commands) {
   return cmdXml;
 };
 
+$.utils.jsonStringifyPhysicals = function(value) {
```
cpcallen: I think it would be much preferable to do the $.physical -> .name conversion as a separate pass from stringification:
* Less chance of introducing error in stringification.
* Easy to swap in replacer-based implementation later.

demo/core_10_base.js
```diff
@@ -202,29 +228,28 @@ $.thing.getCommands = function(who) {
...
-      if (commands.length) {
-        text += '    ' + $.utils.commandMenu(commands) + '\n';
-      }
-      text += '  </' + thing.xmlTag + '>\n';
-    }
+  for (var i = 0; i < contents.length; i++) {
+    var object = contents[i];
+    json.contents.push({
+      type: $.user.isPrototypeOf(object) ? 'user' : 'thing',
+      what: object,
```
cpcallen: The `what` attribute seems like a code smell to me.

demo/core_10_base.js
```diff
@@ -263,22 +288,28 @@ $.room.narrateAll = function(text, obj) {
   }
 };
 
-$.room.tell = function(text) {
+$.room.writeJson = function(json) {
+  if (this.onTell) {
```
cpcallen: As discussed.

demo/core_10_base.js
```diff
@@ -289,16 +320,18 @@ $.user = Object.create($.physical);
...
 
 $.user.say = function(cmd) {
   if (user.location) {
     // Format:  "Hello.    -or-    say Hello.
     var text = (cmd.cmdstr[0] === '"') ? cmd.cmdstr.substring(1) : cmd.argstr;
-    user.location.tellAll(
-        '<say user="' + $.utils.htmlEscape(user.name) + '" room="' +
-        $.utils.htmlEscape(user.location) + '">' +
-        $.utils.htmlEscape(text) + '</say>');
+    var json = {
```
cpcallen: Even though it's arbitrary, I think we should pick a better var name here (and in related places).  We use `cmd` for verb call information, even though it is also plain old JS objects.  How about `activity`, `event` , `update`, `notification`, or even just `msg`?

demo/core_40_world.js
```diff
@@ -28,7 +28,11 @@
   hangout.name = 'Hangout';
   hangout.description = 'A place to hang out, chat, and program.';
   hangout.roll = function(cmd) {
-    this.tellAll('<iframe src="https://www.youtube.com/embed/dQw4w9WgXcQ?autoplay=1"></iframe>')
+    var json = {
+      type: 'iframe',
+      url: "https://www.youtube.com/embed/dQw4w9WgXcQ?autoplay=1"
```
cpcallen: So evil.

www-static/client/common.js
```diff
@@ -67,17 +67,22 @@ CCC.Common.verifyMessage = function(e) {
 
 /**
  * Create a command menu icon.
- * @param {!Element} node Root DOM element describing the menu commands.
+ * @param {!Array<string>|!Element} cmds Array of menu commands,
```
cpcallen: Documentation fixes as discussed.

www-static/client/world.js
```diff
@@ -139,70 +139,60 @@ CCC.World.receiveMessage = function(e) {
...
-    if (dom.getElementsByTagName('parsererror').length) {
-      // Not valid XML, treat as string literal.
-      dom = document.implementation.createDocument(null, 'text');
-      dom.firstChild.appendChild(document.createTextNode(text));
-    }
-    CCC.World.preprocessXml(dom);
-    for (var i = 0, msg; msg = dom.childNodes[i]; i++) {
-      CCC.World.renderMessage(msg);
+    try {
+      var json = JSON.parse(text);
```
cpcallen: Same comment about `json` here as in the core code above.  (Except even more so, since `dom` was at least accurate but `json` is incorrect, because the variable contains POJOs not a JSON string.)  I think this function being called `receiveMessage` suggests `msg` might be the best choice of those I suggested earlier.

www-static/client/world.js
```diff
@@ -139,70 +139,60 @@ CCC.World.receiveMessage = function(e) {
...
-  for (var i = svgTextNodes.length - 1; i >= 0; i--) {
-    var svgTextNode = svgTextNodes[i];
-    var svgNode = CCC.World.stringToSvg(svgTextNode.textContent);
-    if (svgNode) {
-      var container = document.createElement('svgdom');
-      container.appendChild(svgNode);
-      svgTextNode.parentNode.replaceChild(container, svgTextNode);
-    } else {  // Syntax error in SVG.
-      svgTextNode.parentNode.removeChild(svgTextNode);
+CCC.World.preprocessJson = function(json) {
```
cpcallen: Ditto here.

www-static/index.html
NeilFraser: "Budding" will get us into trouble with HR.  "Aspiring" implies that they aren't programmers.  "Inquisitive"?

www-static/client/connect.css
NeilFraser: The JFK standard buttons have a radius on all four corners.  However, the buttons on the connect page are at the top of the screen, so we want the top corners to be square, and leave the bottom corners be radiused.

www-static/index.html
cpcallen: "Budding: (of a person) beginning and showing signs of promise in a particular sphere."  Not sure what the HR problem is.  I think that pretty much sums up exactly our target group.

cpcallen: (But "inquisitive" is O.K. too.)

NeilFraser: That's the third definition.  The second is the problematic one.

PR #101, opened 2017-08-30
Stuff

NeilFraser:


demo/core_10_base.js
```diff
@@ -325,7 +325,7 @@ $.room.tellAll = function(json) {
 $.user = Object.create($.physical);
 $.user.name = 'User prototype';
 $.user.connection = null;
-$.user.svgText = '<circle cx="50" cy="50" r="10" /><line x1="50" y1="60" x2="50" y2="80"/><line x1="40" y1="70" x2="60" y2="70"/><line x1="50" y1="80" x2="40" y2="100"/><line x1="50" y1="80" x2="60" y2="100"/>';
```
cpcallen: [TIL](https://stackoverflow.com/questions/462741/space-before-closing-slash).

demo/core_10_base.js
```diff
@@ -387,7 +387,19 @@ $.user.eval = function($$$cmd) {
       $$$cmd = 'Unhandled exception: ' + String(e);
     }
   }
-  user.narrate($$$cmd);
+  // Save the value without creating a new variable.
+  $.user.eval.value_ = $$$cmd;
```
cpcallen: Maybe this whole function should be restructured as:

```JS
  try {
    (function (result) { /* report normal eval result */ })(eval($$$cmd));
  } catch (e) {
    (function (exception) { /* report abnormal eval result */ })(e);
  }

(This could also allow the result message to more clearly distinguish between a *thrown* Error value and a *returned* one.)

File: demo/core_40_world.js [hunk 3]
```diff
@@ -74,18 +74,26 @@
   };
   clock.moveTo(hangout);
   clock.chime = function(silent) {
+    var minuteMs = 60 * 1000;
+    var hourMs = 60 * minuteMs;
```
cpcallen: I'd have written msPerMinute and msPerHour, just to avoid having Megaseconds in my var name...

www-static/client/log.js
```diff
@@ -372,7 +372,11 @@ CCC.Log.renderJson = function(json) {
       // {type: "narrate", source: "Max", where: "Hangout", text: "Max smiles."}
       // {type: "narrate", source: "Cat", where: "Hangout", text: "Cat meows."}
       var div = document.createElement('div');
-      div.appendChild(document.createTextNode(json.text));
+      var text = json.text;
+      if (json.source) {
+        text = json.source + ': ' + text;
```
cpcallen: I still think this is not the right way to fix the problem—i.e., I think narrations should not in general be associated with objects.  (However, the change is correct in allowing attributed narrations to be displayed meaningfully.)

PR #102, opened 2017-08-31
JSSP

NeilFraser:


demo/core_10_base.js
```diff
@@ -143,32 +169,49 @@ $.physical.moveTo.updateScene_ = function(room) {
...
-  html += '</td></tr></table>';
+  var html = $.jssp.generateOutput(this.generateLook, this);
   user.writeJson({type: "html", htmlText: html});
 };
 $.physical.look.verb = 'l(ook)?';
 $.physical.look.dobj = 'this';
 $.physical.look.prep = 'none';
 $.physical.look.iobj = 'none';
 
+$.physical.generateLook = function(request, response, out) {
```
cpcallen: I'm wondering if we might want a naming convention for JSSP methods.  Perhaps this should be `$.physical.lookJSSP` (or …`lookJssp`, if you prefer)?

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview JavaScript Server Pages for Code City.
+ * @author frase@google.com (Neil Fraser)
```
cpcallen: `fraser`?

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+  if (buffer) {
+    code.push('out.print(' + JSON.stringify(buffer) + ');');
+  }
+  var code = '\n' + code.join('\n') + '\n';
+  var newFunc = new Function('request, response, out', code);
+  $.utils.transplantProperties(func, newFunc);
+  return newFunc;
+};
+
+$.jssp.generateOutput = function(func, thisValue) {
```
cpcallen: Why not `$.jssp.run`?

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+  var request = {};
+  var response = {};
+  var out = new $.jssp.OutputBuffer();
+  func.call(thisValue, request, response, out);
+  return out.toString();
+};
+
+$.jssp.OutputBuffer = function() {
```
cpcallen: Hmm.  This might be our first constructor.  No objection, but different in style to the rest of our core code so far.

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+ * @fileoverview JavaScript Server Pages for Code City.
+ * @author frase@google.com (Neil Fraser)
+ */
+
+$.jssp = {};
+
+$.jssp.compile = function(func) {
+  var src = func.jssp;
+  if (!src) {
+    throw ReferenceError('jssp not found on function.');
```
cpcallen: `'.jssp source property not found`…?

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+  var tokens = src.trim().split(/(<%(?:--|=)?|(?:--)?%>)/);
```
cpcallen: TIL you can capture bits of your separators and splice them into the split result array.  Nice!

But: why not capture the whole <%...%> as a single token?  That would simplify the main loop, because you wouldn't need to maintain any state—you could just examine the first few characters of each token.

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+        } else if (token === '<%=') {
+          state = STATES.EXPRESSION;
+        } else if (token === '<%--') {
+          state = STATES.COMMENT;
+        } else {
+          buffer += token;
+          flush = false;
+        }
+        if (flush && buffer) {
+          code.push('out.print(' + JSON.stringify(buffer) + ');');
```
cpcallen: We should really implement toSource, and then we could get rid of these confusing JSON.stringify calls.

demo/core_32_$.jssp.js
```diff
@@ -0,0 +1,128 @@
...
+          state = STATES.STATEMENT;
+        } else if (token === '<%=') {
+          state = STATES.EXPRESSION;
+        } else if (token === '<%--') {
+          state = STATES.COMMENT;
+        } else {
+          buffer += token;
+          flush = false;
+        }
+        if (flush && buffer) {
```
cpcallen: Is this worth the complexity?  It seems to me that you will only ever get one literal token in a row, so all the buffering etc. seems like a waste of time.  (And even if I'm wrong, in the worst case you'd just get extra calls to out.print(), which is not ideal for performance but hardly a disaster.)

www-static/client/log.js
```diff
@@ -437,7 +437,18 @@ CCC.Log.renderHtmltext = function(div, node) {
       CCC.Log.renderHtmltext(div, node.childNodes[i]);
     }
     if (CCC.Log.renderHtmltext.BLOCK_NAMES.indexOf(node.tagName) !== -1) {
-      if (div.lastChild && div.lastChild.tagName !== 'BR') {
+      // Add a <br> tag, but not if there's already one.
+      var lastTag = div.lastChild;
+      while (lastTag && lastTag.nodeType !== 1) {
+        if (lastTag.nodeType === 3 && lastTag.nodeValue.trim()) {
```
cpcallen: What are these numeric constants `1`, `3`?

PR #103, opened 2017-09-04
Restore socket listeners when recovering from checkpoint

cpcallen:
Plus: encapsulate information about listened ports (and associated net.Server objects)  in new intrp.Server inner class.


cpcallen:
Withdrawing this for now due to it breaking checkpointing in the actual server.

PR #110, opened 2017-09-06
Lost commits, including improvments to pseudoToNative

cpcallen:
* Improve pseudoToNative to handle Errors.
   * Plus copy "extra" properties on Arrays and RegExps, just as for plain
Objects.

* Remove one unnecessary use of arrow function.


cpcallen:
Added tests; removed debugging console.log.  Ready for re-review.

cpcallen:
@NeilFraser: do have a look at this again if you have a chance.

PR #111, opened 2017-09-07
Allow focus to stay on world/log frames until keyboard activity.

NeilFraser:
Allows for copying of text from client.


static/client/common.js
```diff
@@ -169,7 +171,61 @@ CCC.Common.commandFunction = function() {
...
+ */
+CCC.Common.keyDown = function(e) {
+  if (e.key === 'Alt' || e.key === 'Control' || e.key === 'Meta') {
+    // Don't steal focus if the user is pressing a modifier key in preparation
+    // for a cut/copy operation.
+    return;
+  }
+  if (e.ctrlKey || e.altKey || e.metaKey) {
+    // Allow Chrome time to complete a copy before moving focus.
+    setTimeout(CCC.Common.parentFocus, 0);
```
cpcallen: Should `e` be a parameter to `parentFocus` here?  I am thinking in particular of the case that the user is cursoring with Emacs cursor keys (C-f, C-b, C-n, C-p etc.)

demo/core_33_$.pages.js
```diff
@@ -106,13 +106,23 @@ $.pages.home.jssp = [
...
   '  <p id="tagline">A community of inquisitive programmers.</p>',
   '  <iframe src="/login"></iframe>',
+  '  <script>',
+  '  var light = document.getElementById(\'light\');',
+  '  var lightActive = false;',
+  '  function blink() {',
+  '    light.style.fill = lightActive ? \'#f00\' : \'none\';',
+  '    lightActive = !lightActive;',
+  '  }',
+  '  setInterval(blink, 1000)',
```
cpcallen: Disappointed this is not 100% CSS.  Surely you could even do slow transitions, as if it were incandescent?

static/client/common.js
NeilFraser: Tricky.  That means that on Windows C-c, C-x, apparently need to be handled differently than C-f, etc.  Let's revisit this issue when we are together, otherwise I'd be adding untested code and guessing at the results.

PR #116, opened 2017-10-05
Animate the error logo.

NeilFraser:


NeilFraser:
Interesting about the reload.  I had only been testing it on file:// which does a full reload.  But http://localhost does as you describe.

The only solution I've found is to inject the image with a cache-busting param using JS:
<script>
  document.write('');
</script>

What do you think?

cpcallen:
My main thought is that we specifically *want* the image to be cached—particularly on the 502 page which is going to be served a lot if the site is overloaded.

If we can't figure out how to get the SVG animation to restart on page load without breaking cacheing, could the animation be changed to not be unidirectional?

PR #119, opened 2018-02-16
Fixes to disconnection notification.

NeilFraser:


static/client/connect.js
```diff
@@ -147,12 +147,20 @@ CCC.ConnectionStates = {
 CCC.connectionState = CCC.ConnectionStates.NEVER_CONNECTED;
 
 /**
- * Enum for message types.
+ * Enum for message types to the log/world frames.
+ * Should be identical to CCC.Common.MessageTypes
```
cpcallen: We agree that this should be in a separate file.

static/client/connect.js
```diff
@@ -340,16 +348,16 @@ CCC.receiveMessage = function(e) {
 };
 
 /**
- * Distribute a line of text to all frames.
+ * Distribute a line of text to all frames.  If paused, hold this message back.
+ * @param {!CCC.MessageTypes} mode Message type.
  * @param {string} line Text to or from Code City.
```
cpcallen: `text`

PR #121, opened 2018-03-06
Test serialization of `PropertyIterator`

cpcallen:
Including some modifications to `runTest` to allow a fairly aggressive test where we serialize and deseralize between every step of a for-in loop.


cpcallen:
Rebased.  Plz review.

PR #125, opened 2018-03-14
Procrastination

cpcallen:
A bunch of cleanup that I did instead of (well, actually: while) implementing permissions.  :-/

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1023,14 +1023,14 @@ Interpreter.prototype.initArray_ = function() {
   };
   this.createNativeFunction('Array.prototype.splice', wrapper, false);
 
-  wrapper = function(opt_begin, opt_end) {
+  wrapper = function(begin, end) {
     var list = new intrp.Array;
-    var begin = getInt(opt_begin, 0);
+    begin = getInt(begin, 0);
     if (begin < 0) {
       begin = this.properties.length + begin;
     }
     begin = Math.max(0, Math.min(begin, this.properties.length));
-    var end = getInt(opt_end, this.properties.length);
+    end = getInt(end, this.properties.length);
     if (end < 0) {
       end = this.properties.length + end;
     }
@@ -1044,7 +1044,7 @@ Interpreter.prototype.initArray_ = function() {
   };
   this.createNativeFunction('Array.prototype.slice', wrapper, false);
 
-  wrapper = function(opt_separator) {
+  wrapper = function(separator) {
     var cycles = intrp.toStringCycles_;
     cycles.push(this);
     try {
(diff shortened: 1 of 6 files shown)


server/startup/es5.js
```diff
@@ -34,9 +34,6 @@ var decodeURI = new 'decodeURI';
 var decodeURIComponent = new 'decodeURIComponent';
 var encodeURI = new 'encodeURI';
 var encodeURIComponent = new 'encodeURIComponent';
-var suspend = new 'suspend';
-var setTimeout = new 'setTimeout';
-var clearTimeout = new 'clearTimeout';
```
NeilFraser: Removing this should be coupled with a corresponding addition to the core, demo, and minimal cores.

server/interpreter.js
```diff
@@ -1201,18 +1198,17 @@ Interpreter.prototype.initString = function(scope) {
...
   this.createNativeFunction('String.prototype.repeat', wrapper, false);
 };
 
 /**
  * Initialize the Boolean class.
- * @param {!Interpreter.Scope} scope Global scope.
  */
-Interpreter.prototype.initBoolean = function(scope) {
-  var thisInterpreter = this;
+Interpreter.prototype.initBoolean_ = function() {
```
NeilFraser: Functions with a trailing underscore should be marked as @private

server/interpreter.js
```diff
@@ -2209,17 +2196,17 @@ Interpreter.Completion = {
  * @param {Interpreter.Value} value Value to be thrown.  If message is
  *     provided a new error object is created using value as the
  *     prototype; if not it is used directly.
- * @param {string=} opt_message Message being thrown.
+ * @param {string=} message Message being thrown.
```
NeilFraser: {string|undefined}

server/interpreter.js
```diff
@@ -2320,13 +2307,15 @@ Interpreter.prototype.unwind = function(type, value, label) {
 
 /**
  * Class for a scope.
- * @param {Interpreter.Scope=} parentScope Inherited scope.  Defaults to null.
+ * @param {?Interpreter.Scope=} outerScope The enclosing scope ("outer
```
NeilFraser: Is the = valid?

server/interpreter.js
```diff
@@ -2377,12 +2366,12 @@ Interpreter.Thread.prototype.sleepUntil = function(resumeAt) {
 
 /**
  * Returns the original source code for current state.
- * @param {number=} opt_index Optional index in stack to look from.
+ * @param {number=} index Optional index in stack to look from.
```
NeilFraser: |undefined

server/interpreter.js
cpcallen: No.

server/interpreter.js
cpcallen: No.

NeilFraser:
LGTM

PR #129, opened 2018-03-20
Create a REPL for test/debugging

cpcallen:
At last: command-line javascript(-in-javascript)!

### DIFF
--- a/server/repl
+++ b/server/repl
@@ -3,7 +3,7 @@
  * @license
  * Code City: JavaScript Read Eval Print Loop
  *
- * Copyright 2013 Google Inc.
+ * Copyright 2018 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


server/repl
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env node
+/**
+ * @license
+ * Code City: JavaScript Read Eval Print Loop
+ *
+ * Copyright 2013 Google Inc.
```
NeilFraser: Practicing being a Time Lord again?

server/repl
```diff
@@ -0,0 +1,54 @@
...
+});
+
+rl.prompt();
+rl.on('line', function(line) {
+  intrp.value = undefined;
+  intrp.createThread(line);
+  intrp.run();
+  console.log(intrp.pseudoToNative(intrp.value));
+  rl.prompt();
+}).on('close', () => {
```
NeilFraser: Until we start supporting ES6 in Code City, let's not use ES6 where ES5 is just as easy to use.

server/repl
cpcallen: Fixed.

server/repl
cpcallen: Yeah.  Code was copied for documentation for `readline`; I fixed the one above but missed this one.

PR #134, opened 2018-04-01
Exceptions refactor

cpcallen:
Make sure that when a user exception is thrown it will have the correct owner and be handled by the correct thread:

- Actually throw user exceptions.
- Handle unwinding the stack in `step()` and `run()`.
- Ensure that errors created by methods on intrp.Object (and descendants) all use a valid owner, requiring one to be passed in if necessary.
- Make sure that `.thread.perms()` is only called from old-style native functions (and a few other internal methods that will soon go away, e.g. getProperty).

Along the way:

- Abolish `throwError()` in favour of `throw new intrp.Error`
- Convert several more native functions to use `new intrp.NativeFunction`
- Various other minor fixes, `TODO`s (added), etc.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -2857,6 +2857,18 @@ Interpreter.prototype.installTypes = function() {
   };
 
   /**
+   * Add a .name property to this function object.  Partially
+   * implements SetFunctionName from §9.2.11 of the ES6 spec.
+   * @param {string} name Name of function.
+   */
+  intrp.Function.prototype.setName = function(name) {
+    if (Object.getOwnPropertyDescriptor(this.properties, 'name')) {
+      throw Error('Function alreay has name??');
+    }
+    intrp.setProperty(this, 'name', name, Descriptor.c);
+  };
+
+  /**
    * The [[Call]] internal method defined by §13.2.1 of the ES5.1 spec.
    * Generic functions (neither native nor user) can't be called.
    * @param {!Interpreter} intrp The interpreter.
@@ -2936,18 +2948,6 @@ Interpreter.prototype.installTypes = function() {
 
   intrp.UserFunction.prototype = Object.create(intrp.Function.prototype);
   intrp.UserFunction.prototype.constructor = intrp.UserFunction;
-
-  /**
-   * Add a .name property to this function object.  Partially
-   * implements SetFunctionName from §9.2.11 of the ES6 spec.
-   * @param {string} name Name of function.
-   */
-  intrp.Function.prototype.setName = function(name) {
-    if (Object.getOwnPropertyDescriptor(this.properties, 'name')) {
-      throw Error('Function alreay has name??');
-    }
-    intrp.setProperty(this, 'name', name, Descriptor.c);
-  };
 
   /**
    * The [[Call]] internal method defined by §13.2.1 of the ES5.1 spec.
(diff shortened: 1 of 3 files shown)


server/interpreter.js
```diff
@@ -325,11 +305,15 @@ Interpreter.prototype.step = function() {
...
     var nextState = stepFuncs_[node['type']].call(this, stack, state, node);
   } catch (e) {
-    // Eat any step errors.  They have been thrown on the stack.
-    if (e !== Interpreter.STEP_ERROR) {
+    if (e instanceof Error) {
       // Uh oh.  This is a real error in the interpreter.  Rethrow.
       throw e;
+    } else if (typeof e !== 'boolean' && typeof e !== 'number' &&
+        typeof e !== 'string' && e !== undefined && e !== null &&
+        !(e instanceof this.Object)) {
```
NeilFraser: Performance: Reorder these from most- to least-likely.

server/interpreter.js
```diff
@@ -367,12 +351,16 @@ Interpreter.prototype.run = function() {
...
       } catch (e) {
-        nextState = undefined;
-        // Eat any step errors.  They have been thrown on the stack.
-        if (e !== Interpreter.STEP_ERROR) {
+        if (e instanceof Error) {
           // Uh oh.  This is a real error in the interpreter.  Rethrow.
           throw e;
+        } else if (typeof e !== 'boolean' && typeof e !== 'number' &&
+            typeof e !== 'string' && e !== undefined && e !== null &&
+            !(e instanceof this.Object)) {
```
NeilFraser: Same.

server/interpreter.js
```diff
@@ -871,41 +866,62 @@ Interpreter.prototype.initFunction_ = function() {
...
-    if (argsStr) {
-      var args = argsStr.split(/\s*,\s*/);
-      for (var i = 0; i < args.length; i++) {
-        var name = args[i];
-        if (!identifierRegexp.test(name)) {
-          intrp.throwError(intrp.SYNTAX_ERROR,
-              'Invalid function argument: ' + name);
-        }
+  new this.NativeFunction({
+    id: 'Function', length: 1,
```
NeilFraser: Split this line.  I'd be ok with x and y being on the same line, but id and length are unrelated.
Here and several places below.

server/interpreter.js
cpcallen: Hard to say which is more likely when none of them should ever occur unless there is a defect elsewhere in the interpreter.

server/interpreter.js
cpcallen: Thy aren't that unrelated; they are in effect the function signature: `id: Foo, length: 3` means `Foo(arg0, arg1, arg2)`, but without naming the arguments.

server/interpreter.js
cpcallen: I thought about this further, and realised my reply was completely wrong.

The most common case is that the user has thrown a `this.Error`, which is a `this.Object`, so short circuit that case immediately.  Then short circuit if `e === null`, since that's also a valid exception value.  Then invert the other four primitive tests and instead check for `typeof e === 'object' || typeof e === 'function'`, which (having dealt with real `Errors` and all pseudo `Object`s above), shouldn't happen.

PR #136, opened 2018-04-04
Function fixes

cpcallen:
Two fixes to two (unrelated) errors in PR #133.

Now with extra bonus content: Error builtins rewritten as NativeFunctions.


cpcallen:
More bonus content: fix for bug in `Object.prototype.isPrototypeOf` that was causing server to OOM.

PR #139, opened 2018-04-09
Encapsulate .properties

cpcallen:
Encapsulate access to the .properties of interpreter object via methods on Interpreter.prototype.Object whose names and behaviour mirror those of [the JS Proxy handler API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), except:

* All methods take an additional `perms` parameter to specify which user is attempting the operation (for permission-checking purposes, though checks not yet implemented), and
* Only the immediately useful parts of that API are implemented; notably, Object's `.__proto__` is available via `.proto` rather than a call to `.getPrototype`

Also introduce a type `Box`, representing boxed primitives, and an interface `ObjectLike` implemented by both `Box` and `Object` with a subset of the `Object` 'proxy' API, for use in (e.g.) built-in functions that can take either an object or a primitive and are specced as doing a ToObject() on primitives.  Note, however, that the behaviour of Box corresponds with how those functions would behave on actual primitives, rather than boxed primitives, where the two differ—for example, `"foo".bar = 42;` throws `TypeError` rather than performing a non-observable property set on a temporary `String` object.

Finally, rewrite most parts of the interpreter that previously used direct access to `.properties` and/or `.preventExtensions` to instead use the new Object API.  Only the Array built-in methods have not yet been converted.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -688,39 +688,36 @@ Interpreter.prototype.initObject_ = function() {
     id: 'Object.defineProperty', length: 3,
     call: function(intrp, thread, state, thisVal, args) {
       var obj = args[0];
-      var prop = args[1];
-      var descriptor = args[2];
+      var key = args[1];
+      var attr = args[2];
       if (!(obj instanceof intrp.Object)) {
         throw new intrp.Error(state.scope.perms, intrp.TYPE_ERROR,
             'Object.defineProperty called on non-object');
       }
-      prop = String(prop);
-      if (!(descriptor instanceof intrp.Object)) {
+      key = String(key);
+      if (!(attr instanceof intrp.Object)) {
         throw new intrp.Error(state.scope.perms, intrp.TYPE_ERROR,
             'Property description must be an object');
       }
-      if (!obj.properties[prop] && obj.preventExtensions) {
+      if (!obj.properties[key] && obj.preventExtensions) {
         throw new intrp.Error(state.scope.perms, intrp.TYPE_ERROR,
-            "Can't define property '" + prop + "', object is not extensible");
+            "Can't define property '" + key + "', object is not extensible");
       }
       // Can't just use pseudoToNative since descriptors can inherit properties.
-      var nativeDescriptor = {};
-      if (intrp.hasProperty(descriptor, 'configurable')) {
-        nativeDescriptor.configurable =
-            !!descriptor.get('configurable', state.scope.perms);
+      var desc = new Descriptor;
+      if (intrp.hasProperty(attr, 'configurable')) {
+        desc.configurable = !!attr.get('configurable', state.scope.perms);
       }
-      if (intrp.hasProperty(descriptor, 'enumerable')) {
-        nativeDescriptor.enumerable =
-            !!descriptor.get('enumerable', state.scope.perms);
+      if (intrp.hasProperty(attr, 'enumerable')) {
+        desc.enumerable = !!attr.get('enumerable', state.scope.perms);
       }
-      if (intrp.hasProperty(descriptor, 'writable')) {
-        nativeDescriptor.writable =
-            !!descriptor.get('writable', state.scope.perms);
+      if (intrp.hasProperty(attr, 'writable')) {
+        desc.writable = !!attr.get('writable', state.scope.perms);
       }
-      if (intrp.hasProperty(descriptor, 'value')) {
-        nativeDescriptor.value = descriptor.get('value', state.scope.perms);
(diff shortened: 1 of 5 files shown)


server/interpreter.js
```diff
@@ -658,20 +651,22 @@ Interpreter.prototype.initObject_ = function() {
     call: function(intrp, thread, state, thisVal, args) {
       var obj = args[0];
       throwIfNullUndefined(obj);
-      var propsObj = (obj instanceof intrp.Object) ? obj.properties : obj;
-      var names = Object.getOwnPropertyNames(propsObj);
-      return intrp.arrayNativeToPseudo(names, state.scope.perms);
+      if(obj instanceof intrp.Object) {
```
NeilFraser: Space before (.

server/interpreter.js
```diff
@@ -2346,10 +2287,8 @@ Interpreter.prototype.unwind_ = function(type, value, label) {
   if (type === Interpreter.Completion.THROW) {
     // Log exception and stack trace.
     if (value instanceof this.Error) {
-      var name = this.getProperty(value, 'name');
-      var message = this.getProperty(value, 'message');
-      console.log('Unhandled %s: %s', name, message);
-      var stackTrace = this.getProperty(value, 'stack');
+      console.log('Unhandled %s', value.toString());
```
NeilFraser: I think the .toString() is redundant.

server/interpreter.js
```diff
@@ -2734,6 +2859,92 @@ Interpreter.prototype.Server.prototype.unlisten = function(onClose) {
   throw Error('Inner class method not callable on prototype');
 };
 
+/**
+ * @constructor
```
NeilFraser: A high-level comment here about Box would be good.

server/interpreter.js
```diff
@@ -2799,6 +3010,157 @@ Interpreter.prototype.installTypes = function() {
...
+  };
+
+  /**
+   * The [[OwnPropertyKeys]] internal method from ES6 §9.1.12, with
+   * substantial adaptations for Code City including added perms
+   * checks.
+   *
+   * TODO(cpcallen:perms): decide whether null user can read
+   * properties.  (At the moment this is forbidden redundantly by type
+   * siganture an runtime check; one or both should be removed.)
```
NeilFraser: signature

server/interpreter.js
```diff
@@ -3562,66 +3941,160 @@ Interpreter.prototype.installTypes = function() {
   };
 
   /**
+   * Class for a boxed primitive.  Does not @extend
```
NeilFraser: Ah, here's that high-level comment I was looking for earlier.

server/interpreter.js
```diff
@@ -3562,66 +3941,160 @@ Interpreter.prototype.installTypes = function() {
...
    * @constructor
-   * @param {Interpreter.Value} value Value whose properties are to be
-   *     iterated over.
+   * @extends {Interpreter.prototype.Box}
+   * @param {(boolean|number|string)} prim Primitive to box
+   */
+  intrp.Box = function(prim) {
+    /** @private @type {(undefined|null|boolean|number|string)} */
+    this.primitive_ = prim;
+    if(typeof prim === 'boolean') {
```
NeilFraser: Spaces before ( here and twice below.

server/interpreter.js
cpcallen: Done.

server/interpreter.js
cpcallen: Yes, but not in the way you think:

By default, Node's `console.log` prints a nice dump of the structure of the logged object.  The call to `toString()` was intended to avoid dumping the internal structure, so that objects would be shown as `[object Object]`.  But for some reason it doesn't work, and the output actually shows the internal anyway!

server/interpreter.js
cpcallen: As with `Object`, `Function`, etc., the high-level comment is with the actual implementation, at `intrp.Box = ...` in `Interpreter.prototype.installTypes`.

This isn't strictly correct according to the styleguide, but it would have me put the main documentation here and then the implementation would end up with neither the JSDoc comment nor even the type declarations (because the constructor signature is identical).

server/interpreter.js
cpcallen: Done.

server/interpreter.js
cpcallen: No, never mind; I was being an idiot and looking at this code but actually testing the bit a few lines down where non-Error exceptions are dealt with.

Both are now improved.

PR #147, opened 2018-04-28
First-class Threads

cpcallen:
Create a user-visible `Thread` class.

* `setTimeout` becomes a simple polyfilled wrapper around `new Thread(...)`.
* `clearTimeout` becomes a simple polyfilled wrapper around `Thread.kill(...)`.
* `suspend` becomes an alias of `Thread.suspend`.
* `Thread.current()` returns the Thread object for the currently-executing thread.
* `Thread` instances have an `.id` property which is the numeric thread ID.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1956,7 +1956,7 @@ Interpreter.prototype.initThread_ = function() {
   });
 
   new this.NativeFunction({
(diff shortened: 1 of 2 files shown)


cpcallen:
Ping.

PR #149, opened 2018-05-04
Introduce IterableWeakMap

cpcallen:
A subclass of `WeakMap` that implements the entire `Map` API—notably iteration over keys, values, or key-value pairs plus `forEach`, but also `.size` (like `.length` for Arrays) and `.clear()`.

Written in ES6 to new styleguide as it is unavoidably not ES5 compatible and I'd eventually like to package it up as a separate NPM, since there is nothing CodeCity specific about it and we are not the only ones who have use for such a thing.  (Not doing that until we have proved it in production first, though.)

### DIFF
--- a/server/iterable_weakmap.js
+++ b/server/iterable_weakmap.js
@@ -79,14 +79,14 @@ class Cell {
 // closure-compiler supports bounded generic types
 class IterableWeakMap extends WeakMap {
   /**
-    * @param {Iterable<!Array<KEY|VALUE>>|!Array<!Array<KEY|VALUE>>=} iterable
+   * @param {Iterable<!Array<KEY|VALUE>>|!Array<!Array<KEY|VALUE>>=} iterable
    */
   constructor(iterable = undefined) {
     super();
     /** @private @const @type {!Set<!Cell>} */
     this.cells_ = new Set();
 
-    if (iterable == null || iterable === undefined) {
+    if (iterable === null || iterable === undefined) {
       return;
     }
     if (typeof this.set !== 'function') {
--- a/server/iterable_weakmap.js
+++ b/server/iterable_weakmap.js
@@ -1,0 +1,227 @@
+/**
+ * @license
+ * IterableWeakMap
+ *
+ * Copyright 2018 Google Inc.
+ * https://github.com/NeilFraser/CodeCity
+ *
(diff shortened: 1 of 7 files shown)


server/iterable_weakmap.js
```diff
@@ -0,0 +1,227 @@
...
+ * @struct
+ * @implements {Iterable<!Array<KEY|VALUE>>}
+ * @template KEY, VALUE
+ */
+// TODO(cpcallen): Make KEY a bounded to {!Object} once
+// closure-compiler supports bounded generic types
+class IterableWeakMap extends WeakMap {
+  /**
+    * @param {Iterable<!Array<KEY|VALUE>>|!Array<!Array<KEY|VALUE>>=} iterable
+   */
```
NeilFraser: Indent by one space.

server/iterable_weakmap.js
```diff
@@ -0,0 +1,227 @@
...
+  constructor(iterable = undefined) {
+    super();
+    /** @private @const @type {!Set<!Cell>} */
+    this.cells_ = new Set();
+
+    if (iterable == null || iterable === undefined) {
```
NeilFraser: ===

server/iterable_weakmap.js
```diff
@@ -0,0 +1,227 @@
...
+'use strict';
+
+const weak = require('weak');
+
+/**
+ * Function to clean up dead cells from an IterableWeakMap when a key
+ * object has been garbaged collected.
+ * @this {!IterableWeakMap} IterableWeakMap to remove GCed key from.
+ * @param {!Cell} cell Cell containing GCed key.
+ * @return {void}
```
NeilFraser: Remove this?

server/iterable_weakmap.js
cpcallen: Actually, it's the previous line that needs outdenting.

server/iterable_weakmap.js
cpcallen: Fixed.

server/iterable_weakmap.js
cpcallen: I'm going to leave this, for consistency with the official externs for Map (and with some of the declarations earlier in this file.)

I agree that it it is interesting that this is so infrequently used in google3, though.

PR #154, opened 2018-05-07
Fix breakage caused by fd9f0cb

cpcallen:
Missing ')' in cc.js caused SyntaxError at startup.


cpcallen:
Review request NAKed by fraser@.

PR #155, opened 2018-05-07
Use ANYBODY instead of ROOT

cpcallen:
It should not be possible to abuse the (internal) toString methods to obtain informaiton about property values that could not be obtained directly.  Since toString does not (yet?) support a perms argument, use intrp.ANYBODY (isntead of intrp.ROOT) when reading property values to ensure we do not leak information.


cpcallen:
Verbal LGTM from fraser@

PR #160, opened 2018-05-10
Implement Function.prototype.bind natively

cpcallen:
This involves creating a new internal class, `intrp.BoundFunction`, and making a few adjustments to methods on `intrp.Function.prototype`.


server/interpreter.js
```diff
@@ -971,6 +971,34 @@ Interpreter.prototype.initFunction_ = function() {
...
+      var target = thisVal;
+      if (!(target instanceof intrp.Function)) {
+        throw new intrp.Error(state.scope.perms, intrp.TYPE_ERROR,
+            target + ' is not a function');
+      }
+      var thisArg = args[0];
+      var argList = args.slice(1);
+      var perms = state.scope.perms;
+      var f = new intrp.BoundFunction(target, thisArg, argList, perms);
+      var l = 0;
```
NeilFraser: l is a bad variable name.

server/interpreter.js
```diff
@@ -971,6 +971,34 @@ Interpreter.prototype.initFunction_ = function() {
...
+      if (target.has('length', perms)) {
+        var targetLen = target.get('length', perms);
+        if (typeof targetLen === 'number') {
+          l = Math.max(0, targetLen - argList.length);
+        }
+      }
+      f.defineProperty('length', Descriptor.c.withValue(l), perms);
+      var targetName = target.get('name', perms);
+      if (typeof targetName !== 'string') targetName = '';
```
NeilFraser: {}, especially given that the 'number' line above has it.

server/interpreter.js
```diff
@@ -3923,6 +3995,72 @@ Interpreter.prototype.installTypes = function() {
...
+  intrp.BoundFunction = function(func, thisVal, args, owner) {
+    if (!func) return;  // Deserializing
+    intrp.Function.call(/** @type {?} */ (this), owner, func.proto);
+    /** @type {!Interpreter.prototype.Function} */
+    this.boundFunc = func;
+    /** @type {Interpreter.Value} */
+    this.thisVal = thisVal;
+    /** @type {!Array<Interpreter.Value>} */
+    this.args = args;
+
```
NeilFraser: Blank line?

server/interpreter.js
cpcallen: Unfortunately the spec name is *L*, which is also bad as it isn't a constructor.  Changed to `len`.

server/interpreter.js
cpcallen: Done.

server/interpreter.js
cpcallen: There was probably some important code that I erased accidentally, but now I've covered my tracks by removing the `\n` as well.

PR #168, opened 2018-05-17
Editor refinements

NeilFraser:
See individual commits.

### DIFF
--- a/static/code/editor.js
+++ b/static/code/editor.js
@@ -37,6 +37,12 @@ Code.Editor.partsJSON = null;
 Code.Editor.currentEditor = null;
 
 /**
+ * Current source code for editors that haven't yet been created.
+ * @type {string}
+ */
+Code.Editor.uncreatedEditorSource = '';
+
+/**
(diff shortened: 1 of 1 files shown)


demo/core_34_$.www.code.js
```diff
@@ -145,28 +145,35 @@ $.www.code.editor.www = function(request, response) {
...
+      object = $.utils.selector.getGlobal();
+    } else {
+      // Find the origin object.  '$.foo' is the origin of '$.foo.bar'.
+      if (request.parameters.key) {
+        // See if temp ID DB still knows about this key.  If so, we're done.
+        object = $.db.tempId.getObjById(request.parameters.key);
+      }
+      if (!object) {
+        // Parse the parts list.
+        try {
```
cpcallen: This section is all fairly arbitrarily nested.  It might be easier to read if it were structured as follows:
```JS
if (!object && condition1) {
  object = whatever1;
}
if (!object && condition2) {
  object = whatever2;
}
...
```

PR #171, opened 2018-05-22
Error traceback improvements

cpcallen:
Create an (internal) `Thread.prototype.callers` function to get call stack information, and use it as the basis for improving the generation of the `.stack` property in the `intrp.Error` constructor.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -1802,13 +1802,10 @@ Interpreter.prototype.initError_ = function() {
       name: name, length: 1,
       /** @type {!Interpreter.NativeConstructImpl} */
       construct: function(intrp, thread, state, args) {
-        var message = args[0];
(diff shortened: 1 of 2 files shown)


server/interpreter.js
```diff
@@ -4498,46 +4588,46 @@ Interpreter.prototype.installTypes = function() {
...
-        var line = code.substring(lineStart, lineEnd);
-        stack[stack.length] = line;
+      } else if ('eval' in frame) {
+        name = '"' + frame.eval + '"';
+      } else if ('program' in frame) {
+        name = '"' + frame.program + '"';
+      }
+      if ('line' in frame) {
+        line += 'at ' + name + ' ' +
+            String(frame.line) + ':' + String(frame.col);
```
NeilFraser: Is String() needed here?

cpcallen: No.  Removed.

PR #172, opened 2018-05-23
Better Async Functions

cpcallen:
Goal: make it possible to ensure that correct traceback is shown even for errors thrown asynchronously.

At the moment, when an async callback runs while another thread is scheduled (i.e., while `intrp.thread` points at some other thread), and it tries to throw an error, it *will* unwind the correct stack (because the thread to be unwound is stored in a closure in its `reject()` callback), but any `intrp.Error` it creates will get the wrong stack trace because that constructor just uses `intrp.thread` to build its `.stack`.

To fix this, we need to be able to access the `Thread` associated with an async function call from within the body of the function.  The obvious way to do this is the same way we do it for `NativeFunction`s: pass it in as an argument.

One option would be to create an `AsyncNativeFunction` class that would be to `NativeFunction` as `OldAsyncFunction` is to `OldNativeFunction`—i.e., take a `call:` implementation that takes two extra params, being the `resolve` and `reject` functions that will un-block the thread and (in the latter case) unwind the stack.

A more flexible approach, however, is to simply allow any NativeFunction to block (in addition to the various other things they can do already).  To that end:

- Two new `FunctionResult` sentinel values have been added: `Sleep`, which puts the thread to sleep (the function is expected to have already set `runAt` to the desired wake time), and `Block`, which puts the thread to sleep until some external force, such as a an async callback, wakes it up.

- The `resolve`/`reject` mechanism from `intpr.OldAsyncFunction.prototype.call` has been factored out in to a stand-alone generator function `Interpreter.prototype.getResolveReject`.

- All existing `OldAsyncFunction` instances have been rewritten as plain `NativeFunction` instances whose [[Call]] implementations which just happen to return `FunctionResult.Block` at some point.

Note that this does not in fact solve the incorrect-`.stack` problem itself, but does make available the necessary information at the point that `intrp.Error` instances are being created.


server/interpreter.js
```diff
@@ -4317,26 +4317,28 @@ Interpreter.prototype.installTypes = function() {
...
-    if (options.call) {
-      this.call = options.call;
-    }
-    if (options.construct) {
-      this.construct = options.construct;
-    }
-    if (options.name || options.id) {
-      var id = (options.id || options.name);
+    // Register builtin if possible.
+    var id = (options.id || options.name);
```
NeilFraser: Unneeded parens.

server/interpreter.js
```diff
@@ -5430,10 +5433,21 @@ stepFuncs_['Call'] = function (thread, stack, state, node) {
...
+          thread.status = Interpreter.Thread.Status.BLOCKED;
+          return;
+        case Interpreter.FunctionResult.CallAgain:
+          state.step_ = 0;
+          return;
+        case Interpreter.FunctionResult.Sleep:
+          thread.status = Interpreter.Thread.Status.SLEEPING;
+          return;
+        default:
+          throw Error('Unknown FunctionResult??');
```
NeilFraser: This is a great opportunity for an interbang.

server/tests/interpreter_test.js
```diff
@@ -1066,22 +1069,26 @@ exports.testNetworking = async function(t) {
...
-            });
-            client.on('error', function() {
-              reject();
-            });
+    intrp.addVariableToScope(intrp.global, 'recieve', new intrp.NativeFunction({
+      name: 'recieve', length: 0,
+      call: function(intrp, thread, state, thisVal, args) {
+        var reply = '';
+        var rr = intrp.getResolveReject(thread, state);
+        // Recieve some data from the server.
```
NeilFraser: I before E except after C.

server/interpreter.js
cpcallen: Indeed, and they make the intent less clear because they sort of suggest `id` may be being set to a boolean value, which is not at all the case.  Fixed.

server/tests/interpreter_test.js
cpcallen: Fixed.  (So, you *are* looking at the tests…)

server/interpreter.js
cpcallen: Notation somewhat inspired by [Chess annotation symbols](https://en.wikipedia.org/wiki/Chess_annotation_symbols).  "Blunder" is definitely more appropriate than "dubious move" in this case.

PR #176, opened 2018-05-25
Fix .stack generation

cpcallen:
Previously we tried to set `.stack` on Error objects at construction time, in the internal `intrp.Error` constructor.  Unfortunately this depended on using `intrp.thread` to find the current thread, and, outside of `.step` or `.run`—and in particular, in async callbacks—this might be `null` or (worse) point at the wrong thread.

PR #172 attempted to solve this by making the current thread available as a lexical variable in the async function callbacks, but after looking carefully at all the places `Error` objects can be constructed it became clear that this would still leave many possible avenues for an `Error` to end up with the wrong thread's stack in its `.stack`.

So, instead, set `.stack` in the built-in `Error` constructors (which covers all `Error` objects created by user code), and _also_ whenever an `Error` object is thrown (which covers all `Error` objects created by native code).

### DIFF
--- a/server/tests/interpreter_test.js
+++ b/server/tests/interpreter_test.js
@@ -862,12 +862,20 @@ exports.testAsync = function(t) {
       }
   `).thread;
   // Create thread to call async function.
-  var asyncThread = intrp.createThreadForSrc('async()').thread;
+  var asyncThread = intrp.createThreadForSrc('async();').thread;
   intrp.run();
   // asyncThread has run once and blocked; bgThread has run ten times
   // and is now sleeping for 1s.
+
+  // Create Error err to throw.  It should have no stack to start with.
+  var err = new intrp.Error(intrp.ROOT, intrp.ERROR, 'sample error');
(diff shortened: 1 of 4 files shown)


server/tests/interpreter_test.js
```diff
@@ -862,12 +862,20 @@ exports.testAsync = function(t) {
...
   // asyncThread has run once and blocked; bgThread has run ten times
   // and is now sleeping for 1s.
+
+  // Create Error err to throw.  It should have no stack to start with.
+  var err = new intrp.Error(intrp.ROOT, intrp.ERROR, 'sample error');
+  t.assert(name + ': Error has no .stack initially',
+      !err.has('stack', intrp.ROOT));
+
+  // Throw err.
+  intrp.verbose = true;  // REMOVE THIS BEFORE COMMIT
```
cpcallen: Hopefully this was the only mistake you missed in the review, @NeilFraser.

PR #179, opened 2018-05-29
Optimise ConditionalExpression/IfStatement and LogicalExpression

cpcallen:
These two step functions do not need to be revisited once it's been determined which subexpression (sub-statement) will be evaluated, so don't.

Unfortunately this doesn't avoid any allocation, so the performance boost is not as great as it would be if we could avoid creating a whole `State` object, but makes some difference and is a prerequisite to doing proper tail calls.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -5516,25 +5516,19 @@ stepFuncs_['ConditionalExpression'] = function (thread, stack, state, node) {
     state.step_ = 1;
     return new Interpreter.State(node['test'], state.scope);
   }
-  if (state.step_ === 1) {  // Test evaluated; result is in .value
-    state.step_ = 2;
-    var value = Boolean(state.value);
-    if (value && node['consequent']) {
-      // Execute 'if' block.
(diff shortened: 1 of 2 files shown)


server/interpreter.js
```diff
@@ -6000,13 +5993,12 @@ stepFuncs_['ReturnStatement'] = function (thread, stack, state, node) {
  */
 stepFuncs_['SequenceExpression'] = function (thread, stack, state, node) {
   var n = state.n_;
-  var /** ?Interpreter.Node */ expression = node['expressions'][n];
-  if (expression) {
-    state.n_ = n + 1;
-    return new Interpreter.State(expression, state.scope);
+  var /** ?Interpreter.Node */ expression = node['expressions'][n++];
+  if (n >= node['expressions'].length) {
+    stack.pop();
```
NeilFraser: Did we conclude that pop() was faster than length--?

cpcallen: Yes.

PR #183, opened 2018-06-04
Set function names automatically

cpcallen:
ES6 mandates that `Function` objects created from anonymous function expressions have their `name` property set under certain circumstances.  This PR implements all of those cases (except those that involve destructuring assignment), plus an additional case that is especially important to our usage: assigning an anonymous function to a property on an object.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -4830,15 +4830,15 @@ Interpreter.prototype.installTypes = function() {
     }
     var stack = [];
     for (var i = 0; i < callers.length; i++) {
-      var line = '    ';
+      var /** string */ line = '    ';
       var frame = callers[i];
       if ('func' in frame) {
-        var func = frame.func;
-        var name;
+        var /** !Interpreter.prototype.Function */ func = frame.func;
+        var /** string */ name;
         try {
           var pd = func.getOwnPropertyDescriptor('name', perms);
           if (pd) {
-            name = pd.value;
+            name = String(pd.value);
           } else {
             name = 'anonymous function';
           }
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -5307,6 +5307,15 @@ var isIdentifierRef = function(node) {
   return node['type'] === 'Identifier';
 };
 
+/**
+ * Returns true iff node is a MemberExpression.
+ * @param {!Interpreter.Node} node The node to be tested.
+ * @return {boolean} True if node is an identifier.
+ */
+var isMemberRef = function(node) {
+  return node['type'] === 'MemberExpression';
+};
+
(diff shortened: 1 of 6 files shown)


server/interpreter.js
```diff
@@ -4212,8 +4212,9 @@ Interpreter.prototype.installTypes = function() {
     // and we need to audit all use of String() throughout the
     // interpreter (including implicit use inside v8-native
     // functions).
-    return 'function ' + this.get('name', intrp.ANYBODY) +
-        '() { [native code] }';
+    var pd = this.getOwnPropertyDescriptor('name', intrp.ANYBODY);
+    var name = pd ? pd.value : '';
+    return 'function ' + name + '() { [native code] }';
```
NeilFraser: What happens if 'name' is an Object.create(null) or some other value that can't be stringified?  Is that possible?  If it is, Code City will crash here.

cpcallen:
Fixed the infinite-recursion bug in `intrp.Function.prototype.toString` by having it only use the `.name` property if it's already a `string`.

We may want to consider improving `intrp.NativeFunction` to save the initial name of the function so that `intrp.NativeFunction.prototype.toString` could use this when creating a string representaiton of the function, instead of using whatever the current value of `.name` is.  The spec doesn't really mandate one way or the other, and V8 entirely ignores `.name` when `toString`ing functions.

(We already ignore `.name` in `intrp.UserFunction.prototype.toString`, where we simply return the original source with whatever name appears there, if any.)

cpcallen:
There was one other place where I was stringifying functions' `.name`s somewhat indiscriminately.  I don't think there is any way to abuse it to cause infinite recursion (or any other kind of crash), but made some improvements to the typing to force an explicit `String(...)` cast.

cpcallen:
Please have a look at the added commits.

server/interpreter.js
```diff
@@ -4213,7 +4213,7 @@ Interpreter.prototype.installTypes = function() {
     // interpreter (including implicit use inside v8-native
     // functions).
     var pd = this.getOwnPropertyDescriptor('name', intrp.ANYBODY);
-    var name = pd ? pd.value : '';
+    var name = pd && typeof pd.value === 'string' ? pd.value : '';
```
NeilFraser: Can we get some parens around (pd && typeof pd.value === 'string')

server/interpreter.js
```diff
@@ -4830,15 +4830,15 @@ Interpreter.prototype.installTypes = function() {
     }
     var stack = [];
     for (var i = 0; i < callers.length; i++) {
-      var line = '    ';
+      var /** string */ line = '    ';
```
NeilFraser: ?

PR #189, opened 2018-06-13
Fix bug in Scope end bound checking

cpcallen:


cpcallen:
Everything about this is named wrong.  Closing and trying again…

PR #192, opened 2018-06-15
Add tests for Interpreter.Scope

cpcallen:
Several mini-projects to improve how we represent references in the interpreter have all been parked due to performance impact, but hey: at least these tests are worth keeping.


cpcallen:
Actually, looks like I've managed to get the mini-project from which this was salvaged running acceptably quick.

cpcallen:
Actually actually, I'm not sure the other branch is ready yet, so I'm re-opening this one.

PR #195, opened 2018-06-15
Add post-parse "compilation" step

cpcallen:
Walk the AST after it is created, to populate it with various properties that will be used later—most notably, the step function, so no lookups in `stepFuncs_` will be required at interpretation time.

Brings best ever observed performance on `benchFibbonacci10k` to 1811ms (with `testRoundtripScopeRefAndPropIter` disabled).

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -3263,7 +3263,7 @@ Interpreter.State = function(node, scope, wantRef) {
   /** @const @type {!Interpreter.Scope} */
   this.scope = scope;
   /** @const @type {!Interpreter.StepFunction} */
-  this.stepFunc = stepFuncs_[node['type']];
+  this.stepFunc = node['stepFunc'];
   /** @private @const @type {boolean} */
   this.wantRef_ = wantRef || false;
   /** @type {Interpreter.Value} */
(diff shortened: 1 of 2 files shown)


server/interpreter.js
```diff
@@ -451,6 +449,46 @@ Interpreter.prototype.stop = function() {
...
+ * @return {!Interpreter.Node} node Root AST node.
+ */
+Interpreter.prototype.compile_ = function(src, perms) {
+  try {
+    var ast = acorn.parse(src, Interpreter.PARSE_OPTIONS);
+  } catch (e) {  // Acorn threw a SyntaxError.  Rethrow as a trappable error?
+    throw perms ? this.errorNativeToPseudo(e, perms) : e;
+  }
+
+  (function analyse(node) {
```
NeilFraser: Is there some reason this isn't a private helper function?  As it stands, this function is being recreated over and over, every time compile_ is called.

cpcallen: Two reasons:
* I don't expect it to be called from anywhere else (and this way it is *actually* private), and
* I foresee future changes for which having access to local variables in `.compile_` will be useful.  In particular, much of the tree-walking work presently done by `.populateScope_` can actually be done upfront in `.compile_`—so it only needs to be done once, instead of once per invocation of a function—and for that it will be convenient to have access to an array to accumulate `var` bindings and so on.

PR #197, opened 2018-06-18
Make the base of an environment reference be the Scope object containing the referee

cpcallen:
Until now, references have been stored as [_base_, _name_] pairs, where _base_ is an object (or primitive) in the case of a property reference or the `SCOPE_REFERENCE` sentinel value in the case of an environment reference.

The spec describes environment references as having their _base_ value be the environment record (i.e., `Scope` object) in which _name_ is actually found; the search through the linked list of environment records (i.e., `Scope` objects) is done at the time the reference is created rather than when it is used, and unresolvable references are represented as having _base_ be undefined.

This PR changes our reference values to accord with the spec.  This has a few benefits:
* It isn't necessary to pass the current innermost `Scope` to `getValue` and `setValue`.
* Evaluating an expressions like `n++` and `a += b` does not entail searching for the variable being updated twice.
* We can eliminate `SCOPE_REFERENCE` and the `Sentinel` class (which also simplifies the serializer).

This change appears to be performance-neutral for the `benchFibbonacci10k` benchmark (or at least within the noise margin).

This work also lead to the discovery of a bunch of problems with property references with scalar bases and MemberExpression which were fixed in PR #191.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -2982,21 +2982,6 @@ Interpreter.FunctionResult.CallAgain = new Interpreter.FunctionResult;
 Interpreter.FunctionResult.Sleep = new Interpreter.FunctionResult;
 
 /**
- * Class for unique sentinel values passed to various functions.
- * Declared so that sentinel values can have a specific type that we
- * can type-check against (though as they share a single type you
- * could still pass the wrong sentinel value to a function).
- * @constructor
- */
-Interpreter.Sentinel = function Sentinel() {};
-
-/**
- * Unique sentinel for indicating that a reference is a variable on the scope,
- * not an object property.
- */
-Interpreter.SCOPE_REFERENCE = new Interpreter.Sentinel();
-
-/**
  * Interpreter statuses.
  * @enum {number}
  */
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -2724,36 +2724,43 @@ Interpreter.prototype.isUnresolvableReference = function(scope, ref, perms) {
 };
 
 /**
- * Gets a value from the scope chain or from an object property.
+ * Gets the value of a referenced name from the scope or object referred to.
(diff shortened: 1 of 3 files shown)


server/interpreter.js
```diff
@@ -2720,45 +2720,45 @@ Interpreter.prototype.calledWithNew = function() {
...
-  if (ref[0] === Interpreter.SCOPE_REFERENCE) {
-    // A null/varname variable lookup.
-    this.setValueToScope(scope, ref[1], value);
-  } else {
-    // An obj/prop components tuple (foo.bar).
-    this.toObject(ref[0], perms).set(ref[1], value, perms);
+Interpreter.prototype.setValue = function(ref, value, perms) {
+  var base = ref[0];
+  var name = ref[1];
+  if (base === null) {  // Unresolvable reference
```
NeilFraser: .

server/interpreter.js
```diff
@@ -3173,6 +3158,57 @@ Interpreter.Scope.prototype.createImmutableBinding = function(name, value) {
...
+ * equivalently the ResolveBinding specification function from ES6
+ * §8.3.1.
+ * @param {string} name Name of variable.
+ * @return {?Interpreter.Scope} The scope that binds name, or null if none.
+ */
+Interpreter.Scope.prototype.resolve = function(name) {
+  for (var s = this; s; s = s.outerScope) {
+    if (name in s.vars) break;
+  }
+  return s;
```
NeilFraser: Might be clearer to say:
```
  if (name in s.vars) return s;
}
return null;
```

server/interpreter.js
cpcallen: Fixed.

server/interpreter.js
cpcallen: Agreed.  Changed.

PR #202, opened 2018-07-05
Formalise builtins registry

cpcallen:
Database dumper will need to be able to look up the names of builtins, so create a Registry class providing bidirectional mapping and use it for .builtins_ (which loses the underscore, because it will need to be public).


cpcallen:
Hold off on reviewing this for a moment; I have another commit coming.

PR #206, opened 2018-07-10
Fix client escaping and add auto-hyperlinking.

NeilFraser:


static/client/common.js
```diff
@@ -287,3 +287,15 @@ CCC.Common.createSvgElement = function(name, attrs, opt_parent) {
...
 };
+
+/**
+ * Given plain text, encode spaces and tabs such that HTML won't crush it.
+ * Does not handle line breaks in any way.
+ * @param {string} text Plain text.
+ * @return {string} HTML with any runs of spaces encoded.
+ */
+CCC.Common.escapeSpaces = function(text) {
+  return text.replace(/\t/g, '\u00A0 \u00A0 ')
```
cpcallen: Tabs are conventionally eight spaces, not four.  (That's leaving aside the issue of actual tabular alignment.)

Also: wouldn't `&nbsp;` be clearer?

static/client/common.js
```diff
@@ -299,3 +299,53 @@ CCC.Common.escapeSpaces = function(text) {
       .replace(/  /g, '\u00A0 ').replace(/  /g, '\u00A0 ')
       .replace(/^ /gm, '\u00A0');
 };
+
+/**
+ * Detect URLs in the provided document fragment and replace with links.
+ * @param {!Element} el A DOM element to scan and modify.
+ */
+CCC.Common.autoHyperlink = function(el) {
+  var isSvg = el.namespaceURI == CCC.Common.NS;
```
cpcallen: `===`?

static/client/common.js
```diff
@@ -299,3 +299,53 @@ CCC.Common.escapeSpaces = function(text) {
...
+  var children = el.childNodes;  // This is a live NodeList.
+  for (var i = children.length - 1, child; (child = children[i]); i--) {
+    if (child.nodeType === 3) {
```
cpcallen: * `Node.TEXT_NODE`.

* Why not move this test up to the top of the function, so the two loops can be separated?:

```JS
  if (el.nodeType !== Node.TEXT_NODE) {
    // split text and loop over parts.
  } else {
    for (child of …) {
      CCC.Common.autoHyperlink(child);
    }
  }
```

* Failing that, at least invert the test and use continue, to avoid unnecessary indentation:
```JS
  if (child.nodeType !== Node.TEXT_NODE) {
      CCC.Common.autoHyperlink(child);
      continue;
    }
    var text = …
```

static/client/common.js
```diff
@@ -299,3 +299,53 @@ CCC.Common.escapeSpaces = function(text) {
...
+        }
+        el.removeChild(child);
+      }
+    } else {
+      CCC.Common.autoHyperlink(child);
+    }
+  }
+};
+
+CCC.Common.autoHyperlink.urlRegex = /((?:https?:\/\/|www\.)[-\w.~:\/?#\[\]@!$&'()*+,;=%]+)/i;
```
cpcallen: Line length.  Wrap at `=`?

Also: is it correct to exclude unicode from URLs?  For ones we generate it can presumably be escaped correctly, but if this will mostly be used for auto-highlighting user-written URLs we might want to be a bit more accepting her.

static/client/common.js
```diff
@@ -299,3 +299,53 @@ CCC.Common.escapeSpaces = function(text) {
...
 };
+
+/**
+ * Detect URLs in the provided document fragment and replace with links.
+ * @param {!Element} el A DOM element to scan and modify.
+ */
+CCC.Common.autoHyperlink = function(el) {
+  var isSvg = el.namespaceURI == CCC.Common.NS;
+  var children = el.childNodes;  // This is a live NodeList.
+  for (var i = children.length - 1, child; (child = children[i]); i--) {
```
cpcallen: Presumably you are processing this in reverse order because `children` is live.  But since neither `i` nor `children` are otherwise referenced, why not
```JS
  for (child of el.childNodes) {
```
(or, if necessary, `for (child of Array.from(el.childNodes))`)?

static/client/common.js
```diff
@@ -299,3 +299,53 @@ CCC.Common.escapeSpaces = function(text) {
...
+    if (child.nodeType === 3) {
+      var text = child.nodeValue;
+      var parts = text.split(CCC.Common.autoHyperlink.urlRegex);
+      if (parts.length > 1) {
+        for (var j = 0; j < parts.length; j++) {
```
cpcallen:
```JS
      if (parts.length <= 1) continue;  // No hyperlinks found.
      for (…
```

static/client/common.js
NeilFraser: Changed to 8.

&nbsp; is HTML (suitable for .innerHTML assignment), we need the equivalent Unicode character (suitable for document.createTextNode useage).

static/client/common.js
NeilFraser: Fixed.

static/client/common.js
NeilFraser: Wrapped.

We'll deal with Unicode later.  Makes the regex way more complicated (full URL parsing regexes are known to crash browsers).  Besides, the failure case here is just a lack of auto hyperlinking in the client.

static/client/common.js
NeilFraser: Done.

static/client/common.js
NeilFraser: Changed in 44 places, but not this one.

static/client/common.js
NeilFraser: Changed to use Node properties.  Though this is not the JS way, as seen in every one of our dependencies.

The first option isn't a great idea since we are passing a text element, but its parent node is the one being modified.  Unexpected side effects.  I'll go with the second option.

NeilFraser:
Ping?  Given that the various people on style group said:
a) use TypeScript
b) don't use for/of since nobody can figure out what it does
c) use Array.from
d) use (the existing) normal loop
I think this PR is fine.

cpcallen:
I was  waiting for you to submit a revised version, since I think the feedback was pretty clearly in favour of for-of overall.  (I actually posted to javascript@, since javascript-style@ is mostly CL notifications, which might explain the incomprehension.)

I don't seem to have reviewed the other changes, so let me do that.  I will not hold up the  PR based on the outstanding issue, but I think I will seek further clarification from the stylists..

NeilFraser:
> I think the feedback was pretty clearly in favour of for-of overall.

Are we looking at the same conversation?

* FR mumbled something about TypeScript.
* JS said to use for-of, then realized his mistake (confirmed by JP) and suggested to use Array.from.
* JZ suggested using a traditional loop.
* SH suggested using nextSibling.

The feedback doesn't appear to be clearly in favour of any option.  Rather it seems to be spread across the board.

demo/core_14_$.utils.html.js
```diff
@@ -0,0 +1,40 @@
...
+$.utils.html = {}
+
+$.utils.html.escape = function(text) {
+  // Escape text so that it is safe to print as HTML.
+  return String(text).replace(/&/g, '&amp;').replace(/"/g, '&quot;')
+      .replace(/</g, '&lt;').replace(/>/g, '&gt;');
+};
+
+$.utils.html.preserveWhitespace = function(text) {
+  // Escape text so that it is safe and preserves whitespace formatting as HTML.
```
cpcallen: I think it might be worth adding a bit more explanation here.  I was puzzling about the two `/  /g`s below, wondering if there were some characters in them which were not displaying (e.g., if they were different kinds of whitespace characters or something).  And worse: I had already looked at this code before, and figured it out, and *still* didn't immediately understand what was going on.

static/client/world.js
```diff
@@ -173,10 +173,12 @@ CCC.World.receiveMessage = function(e) {
  */
 CCC.World.preprocessMessage = function(msg) {
   if (Array.isArray(msg)) {
-    for (var i = 0; i < msg.length; i++) {
-      CCC.World.preprocessMessage(msg[i]);
+    for (var singleMsg of msg) {
```
cpcallen: I thought `singleMsg` was a big ugly, and wondered, given the new block scoping rules as well as the order that things get evaluated in, if it would be legal to write:

```JS
    for (const msg of msg) {
```

…where the first `msg` was a new const binding scoped to the for-of, while the second one was the existing var binding.  But (perhaps fortunately!) this isn't legal: the second `msg` is in fact the same block-scoped const binding as the first, which (since the it has not yet been initialised) is in the temporal dead zone and so you'd get `ReferenceError: msg is not defined`.

I am somewhat relieved.

demo/core_14_$.utils.html.js
NeilFraser: Comment added:
Runs of three spaces ('   ') need to be escaped twice ('_  ', '__ ').

PR #209, opened 2018-07-11
Set RegExp and Date value at construction time

cpcallen:
Since `Date` and `RegExp` object are immutable in ES5.1, set their internal values in the constructor.

I've also inlined the former `intrp.RegExp.prototype.populate` method into the `intrp.RegExp` constructor—but I note that, in ES6, `RegExp.prototype.compile` was added to Appendix B, making RegExps internal state mutable.  If we intend to support that then we should probably keep `.populate` as its own method for now.


cpcallen:
Ping @NeilFraser.

PR #210, opened 2018-07-12
Modify eval to handle ambiguous object literals.

NeilFraser:


demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
     throw e;
   }
 };
+
+$.utils.code.formatForEval = function(src) {
```
cpcallen: "Format" suggests text output formatting.  Maybe `rewriteForEval`?

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+  // Eval treats {} as an empty block (return value undefined).
+  // Eval treats {'a': 1} as a syntax error.
+  // Eval treats {a: 1} as block with a labeled statement (return value 1).
+  // Detect these cases and enclose in parenthesis.
+  // But don't mess with: {var x = 1; x + x;}
+  // This is consistent with the console on Chrome and Node.
+  try {
+    var ast = $.utils.acorn.parse(src, { ecmaVersion: 5 });
```
cpcallen: This set of options is already simplified compared to what we actually pass to acorn, and will become even more so as we begin to implement ES6 features.

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+  } catch (e) {
+    try {
+      ast = $.utils.acorn.parseExpressionAt(src, 0, { ecmaVersion: 5 });
+      if (src.substring(ast.end).trim() !== '') {
+        throw SyntaxError('Not just an expression');
```
cpcallen: Does this reject trailing comments on object literals?  Maybe add some test cases for that, to be sure.

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+        ast.body[0].body[0].type === 'LabeledStatement' &&
+        ast.body[0].body[0].body.type === 'ExpressionStatement') {
+      // This is an object literal: {a: 1}
+      return '(' + src + ')';
+    }
+  }
+  return src;
+};
+
+$.utils.code.formatForEval.test = function() {
```
cpcallen: Maybe `__test__` or something a bit more distinctive?  After all, there is already a `RegExp.prototype.test` which is not unit tests for `RegExp.prototype`.

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+  var matrix = {
```
cpcallen: Convention in the server is  to call this `cases`.

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+    '1 + 2': '1 + 2',
+    '1 + 2;\n3 + 4;': '1 + 2;\n3 + 4;',
+    '{}': '({})',
+    '{};\n3 + 4;': '{};\n3 + 4;',
+    '{"a": 1}': '({"a": 1})',
+    '{"a": 1};\n3 + 4;': '{"a": 1};\n3 + 4;',
+    '{a: 1}': '({a: 1})',
+    '{a: 1};\n3 + 4;': '{a: 1};\n3 + 4;'
```
cpcallen: Some additional test cases:
```JS
'{a: 1}  // Comment';
'{a: 1, b: 2}',
```

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+  };
+  for (var key in matrix) {
```
cpcallen: Add `hasOwnProperty` check here.  After all, in-core `Object.prototype` very likely to accidentally get enumerable properties on occasion; don't want tests breaking (unless they are specifically testing that, of course).

demo/core_13_$.utils.code.js
```diff
@@ -136,3 +136,60 @@ $.utils.code.toSourceSafe = function(value) {
...
+    var actual = $.utils.code.formatForEval(key);
+    if (actual !== matrix[key]) {
+      throw Error('Expected: ' + matrix[key] + ' Actual: ' + actual);
```
cpcallen: Convention in server tests is always "got:" then "want:".

demo/core_13_$.utils.code.js
NeilFraser: How about `unittest`?  I'm not really a fan of adding more and more underscores for uniqueness.

demo/core_13_$.utils.code.js
NeilFraser: From my experience with DMP, the convention is first/second in Python:
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual

 and actual/expected everywhere else:
https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html
https://www.mathworks.com/help/matlab/ref/matlab.unittest.qualifications.assertable.assertequal.html
https://phpunit.de/manual/6.5/en/appendixes.assertions.html
https://wiki.servoy.com/display/DOCS/JSUnit#JSUnit-assertEquals(expected,actual)

I don't mind what the server does, but teaching the actual/expected terminology to our users sounds like a better plan than inventing new terminology for no reason.

demo/core_13_$.utils.code.js
cpcallen: Ahh.  Whereas I would prefer to avoid a name that does not have underscores, least it conflict with some purpose other than unit testing.  I am specifically thinking along the lines of having unit tests run automatically—a situation in which you do not want random, similarly-named functions being called.

(The correct solution would be a unique Symbol, of course…)

demo/core_13_$.utils.code.js
cpcallen: My comment was more about order than terminology.

NeilFraser: JUnit, JSUnit, NUnit, PHPUnit all use expected/actual in that order.  The only package I can find with the reverse order is Matlab.  Sounds like the server's tests ought to be reversed.

demo/core_13_$.utils.code.js
NeilFraser: Done.

demo/core_13_$.utils.code.js
NeilFraser: Done.

demo/core_13_$.utils.code.js
NeilFraser: Done.

demo/core_13_$.utils.code.js
NeilFraser: Done.

demo/core_13_$.utils.code.js
NeilFraser: Changed the default options.

demo/core_13_$.utils.code.js
NeilFraser: Heh, good catch.  Comments don't show up in the AST at all.  So we got back:
`({} // comment)`
which is not valid.  Changed to always add a line break before the closing paren.

demo/core_13_$.utils.code.js
cpcallen: Server's tests were based on the original golang ones, and the Go style guide specifies the order explicitly.  Given that Go appears to be the odd one out I'm happy to have the in-core tests conform to the usual order and terminology for JS.

I may consider changing the output of the server tests, though I think that the Go stylists actually made the right call here, because it makes it easy to have super-meaningful test output like `Fail: factorial(3) == 5 (want 6)`, which is quite a bit more clear than `Fail: testFactorialThree: expected: 6, actual: 5`.

demo/core_13_$.utils.code.js
cpcallen: I think it would be preferable to dispose of the unexpected case at the top and avoid indenting the main body of the loop more than necessary:

```JS
  for ( ... ) {
    if (!cases.hasOwnProperty(key)) continue;
    var actual = ...
```

demo/core_13_$.utils.code.js
```diff
@@ -157,39 +157,44 @@ $.utils.code.formatForEval = function(src) {
...
-    '{"a": 1}': '({"a": 1})',
-    '{"a": 1};\n3 + 4;': '{"a": 1};\n3 + 4;',
-    '{a: 1}': '({a: 1})',
-    '{a: 1};\n3 + 4;': '{a: 1};\n3 + 4;'
+    '{"a": 1}': '({"a": 1}\n)',
+    '{"a": 1};': '{"a": 1};',
+    '{a: 1}': '({a: 1}\n)',
+    '{a: 1};': '{a: 1};',
+    '{a: 1}  // Comment': '({a: 1}  // Comment\n)',
+    '{a: 1, b: 2}': '({a: 1, b: 2}\n)',
```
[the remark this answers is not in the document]
NeilFraser: That's what it already is.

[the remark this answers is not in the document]
NeilFraser: Done.

demo/core_13_$.utils.code.js
NeilFraser: Done

demo/core_13_$.utils.code.js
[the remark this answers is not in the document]
NeilFraser: Done.

PR #212, opened 2018-07-28
New Favicon

NeilFraser:
Replace the temporary 'CC' favicon that looks like a logo for a perfume company, with an icon based on our skyline.  Old vs new:

screen shot 2018-07-23 at 17 35 10

This favicon is made up of three different images:

screen shot 2018-07-23 at 17 35 53


NeilFraser:
The light does not currently blink.  Apparently Firefox does support animated gif favicons (though I can't find a live example), but Chrome does not.  There is a work-around that allows JavaScript to dynamically replace the favicon if we really want that animation, though this JS would have to be on every page.

cpcallen:
I like the fact that you checked…

I don't think we actually want a blinking favicon.  But it would be amusing.

PR #213, opened 2018-07-30
Reduce scene spam

NeilFraser:
Currently if you drop Fido, the server sends two separate messages.  The first is an unrequested scene update that adds Fido to the room.  The second is a room naration of “You drop Fido”.  This renders as two panels, where only one is needed.

### DIFF
--- a/static/client/world.js
+++ b/static/client/world.js
@@ -229,6 +229,38 @@ CCC.World.renderMessage = function(msg) {
     CCC.World.panoramaMessages.push(msg);
     return;
   }
+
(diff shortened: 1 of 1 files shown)


NeilFraser:
The unrequested scene panel displays the room description.  That's a narration.  We can merge narrations from the same source, but not currently from different sources.

Another solution would be to drop the room description from unrequested scene panels.  However, if there's a command that picks up and/or drops a dozen objects in the room, we don't want a dozen panels being drawn.

PR #214, opened 2018-08-01
Correct ‘ID’ cookie capitalization

NeilFraser:
This is the result of a partial rename while attempting to comply with Google cookie style guide.

However, the reason this wasn't detected earlier is troubling.  The 'ID'/'id' cookie is supposed to be a session cookie.  But in practice it lives forever in Chrome and Firefox.  That's because these browsers have redefined 'session' to be perpetual.  When the browsers restart, they resume where they left off, restoring all state.  A bit like Code City, for that matter.

This isn't really our problem, but it is interesting.

### DIFF
--- a/connect/connectServer.js
+++ b/connect/connectServer.js
@@ -129,7 +129,7 @@ function handleRequest(request, response) {
         cookieList[parts.shift().trim()] = decodeURI(parts.join('='));
     });
     // Decrypt the ID to ensure there was no tampering.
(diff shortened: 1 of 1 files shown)


cpcallen:
For posterity: some [commentary on the session cookie fracas](http://blog.petersondave.com/cookies/Session-Cookies-in-Chrome-Firefox-and-Sitecore/) (with links to the Chrome and Firefox bug trackers where the new behaviour has been repeatedly reported as a defect).

PR #215, opened 2018-08-02
Add ‘inspect’ command to every $.physical

NeilFraser:
Opens the code editor.

Attempts to reuse existing windows were not successful.  At issue is that if a window is already open (as a tab), there is no way to switch to that tab from JS.  The API window.focus() doesn't function in Chrome or Firefox.  Thus 'inspect' would just do nothing and the user would have to go hunting for the right tab.

Thus I'm just opening a fresh window every time.

### DIFF
--- a/demo/core_10_base.js
+++ b/demo/core_10_base.js
@@ -210,8 +210,20 @@ $.physical.lookJssp.jssp = [
   '  </tr>',
   '</table>'].join('\n');
 
+$.physical.inspect = function(cmd) {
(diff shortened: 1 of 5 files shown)


NeilFraser:
My assumption at the moment is that users won't be allowed to generate and send arbitrary JSON to the client.  That would allow them to forge content.  Instead they will have access to controlled functions that generate the JSON.

But database security won't exist until the server has its own security model.

PR #216, opened 2018-08-02
Move eval call out of eval verb.

NeilFraser:
Simplify code by isolating the eval() call in a minimal function.

### DIFF
--- a/demo/core_10_base.js
+++ b/demo/core_10_base.js
@@ -402,56 +402,58 @@ $.user.think.dobj = 'any';
 $.user.think.prep = 'any';
 $.user.think.iobj = 'any';
 
-$.user.eval = function($$$cmd) {
+$.user.eval = function(cmd) {
   // Format:  ;1+1    -or-    eval 1+1
-  // To reduce the likelihood of clashes with identifiers in the evaled
-  // code, this function has only a single, awkwardly-named parameter
-  // and has no local variables of its own.  Conversely, however, we
-  // create a few local variables with short, convenient names as
-  // aliases for commonly-used values.
-  $$$cmd =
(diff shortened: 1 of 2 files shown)


demo/core_10_base.js
```diff
@@ -402,57 +402,57 @@ $.user.think.dobj = 'any';
...
   }
-  user.narrate('⇒ ' + $$$cmd);
+  user.narrate('⇒ ' + out);
 };
 $.user.eval.verb = 'eval|;.*';
 $.user.eval.dobj = 'any';
 $.user.eval.prep = 'any';
 $.user.eval.iobj = 'any';
 
+$.user.eval.doEval_ = function($$$src, me, here) {
```
cpcallen: It seems to me that this is one place where we would really like to be able to use `with`—i.e.:

```JS
$.user.eval.doEval_ = function($$$src, $$$env) {
  with ($$$env) {
    return eval($$$src);
  }
};

// Call using:
$.usr.eval.doEval_(src, {me: this, here: this.location});
```
I think it will make sense to provide a version of `eval` that lets the caller supply the scope to use.  For the moment, though, this implementation is fine.

demo/core_10_base.js
```diff
@@ -402,57 +402,57 @@ $.user.think.dobj = 'any';
...
 $.user.think.iobj = 'any';
 
-$.user.eval = function($$$cmd) {
+$.user.eval = function(cmd) {
   // Format:  ;1+1    -or-    eval 1+1
-  // To reduce the likelihood of clashes with identifiers in the evaled
-  // code, this function has only a single, awkwardly-named parameter
-  // and has no local variables of its own.  Conversely, however, we
-  // create a few local variables with short, convenient names as
-  // aliases for commonly-used values.
```
cpcallen: This note (or an appropriately-edited version of it) should be preserved in `doEval_`.

PR #217, opened 2018-08-04
Eval ambiguous expressions

NeilFraser:


demo/core_10_base.js
```diff
@@ -418,7 +418,7 @@ $.user.think.iobj = 'any';
 $.user.eval = function(cmd) {
   // Format:  ;1+1    -or-    eval 1+1
   var src = (cmd.cmdstr[0] === ';') ? cmd.cmdstr.substring(1) : cmd.argstr;
-  src = $.utils.code.rewriteForEval(src);
+  src = $.utils.code.rewriteForEval(src, false);
```
cpcallen: I think it would be fine to make forceExpression optional (i.e., document as such) and omit an explicit false here.

Alternatively, the style guide now (as of March) notes that you may add a comment to make the purpose of the second parameter more clear: `rewriteForEval(src, /* forceExpression= */ false)`.

demo/core_13_$.utils.code.js
```diff
@@ -137,64 +137,111 @@ $.utils.code.toSourceSafe = function(value) {
...
     if (!cases.hasOwnProperty(key)) continue;
-    var actual = $.utils.code.rewriteForEval(key);
-    if (actual !== cases[key]) {
-      throw Error('Expected: ' + cases[key] + ' Actual: ' + actual);
+    for (var i = 0; i < 1; i++) {
+      // Alternate between expression and statement evals.
+      try {
+        actual = $.utils.code.rewriteForEval(key, i === 0);
+      } catch (e) {
+        actual = SyntaxError;
```
cpcallen: So, a recent-ish episode of TotT strongly discourages table-driven tests where the test loop itself (rather than the function under test) does something different depending on the test input—and in particular, that for table-driven tests advocated for separating out the valid and invalid tests into separate tests.  The valid cases will have a table including input and expected output; the invalid test table can have just the invalid inputs.

I can see that in this case it is desirable to avoid repeating the test data, but the resulting deeply nested driver code is kind of ugly and not that easy to follow.

demo/core_13_$.utils.code.js
```diff
@@ -137,64 +137,111 @@ $.utils.code.toSourceSafe = function(value) {
...
-      // This is an object literal: {a: 1}
-      return '(' + src + '\n)';
+    return src;
+  }
+  // Try parsing src as an expression.
+  // This may throw.
+  ast = $.utils.acorn.parseExpressionAt(src, 0);
+  var remainder = src.substring(ast.end).trim();
+  if (remainder !== '') {
+    // Trim off any trailing comments.
```
cpcallen: This comment seems out of place, and isn't the most obvious description of what is happening below.

demo/core_34_$.www.code.js
```diff
@@ -179,8 +179,10 @@ $.www.code.editor.www = function(request, response) {
       if (request.parameters.src) {
         var ok = true;
         try {
-          // TODO(fraser): Make this secure -- somehow.
```
cpcallen: Did you mean to remove this TODO?

demo/core_10_base.js
NeilFraser: I quite like the second option.  Done.

demo/core_13_$.utils.code.js
NeilFraser: Agreed about the second loop.  I've unrolled the loop, which I think makes it easier to follow, though does introduce repeated code.  Amusingly, this was my original code, but I refactored it into a loop because I assumed I'd get yelled at in the code review.

demo/core_13_$.utils.code.js
NeilFraser: Fixed.

demo/core_34_$.www.code.js
NeilFraser: That TODO basically applies to the entirety of the database, so it's a bit misleading on that specific line.

demo/core_13_$.utils.code.js
```diff
@@ -231,17 +232,25 @@ $.utils.code.rewriteForEval.unittest = function() {
...
-      }
-      if (actual !== cases[key][i]) {
-        throw Error((i ? 'Eval Statement' : 'Eval Expression') +
-                    '\nExpected: ' + cases[key][i] + ' Actual: ' + actual);
-      }
+    // Test eval as an expression.
+    try {
+      actual = $.utils.code.rewriteForEval(key, true);
+    } catch (e) {
+      actual = SyntaxError;
```
cpcallen: I should create a TC-30 proposal to add `` `!' `` syntax to JavaScript…

demo/core_13_$.utils.code.js
cpcallen: Yeah: the longer version is definitely more readable but the whole try/catch business is now extra-obnoxious.  I don't know if there's a winning plan here.

PR #221, opened 2018-08-10
Styleguide updates

cpcallen:
There have been a few small changes in the JS style guide in the last year.  Implement them in the server codebase.

(Note that we are still not compliant with the styleguide in general: e.g. most of the server is ES5…)


NeilFraser:
LGTM

PR #223, opened 2018-08-13
Fix bug causing setPrototypeOf(..., null) to crash

cpcallen:
If new prototype is null then setPrototype woudl fail because null has no .properties property.  Fix this, and a separate typo bug which was causing some DB tests to be silently skipped.

### DIFF
--- a/server/tests/testcases.js
+++ b/server/tests/testcases.js
@@ -1223,6 +1223,14 @@ module.exports = [
     `,
     expected: 'p' },
 
+  { name: 'Object.setPrototypeOf(..., null)', src: `
+    var o = {parent: 'o'};
+    var q = Object.create(o);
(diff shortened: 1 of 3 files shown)


server/tests/testcases.js
```diff
@@ -1223,6 +1223,14 @@ module.exports = [
     `,
     expected: 'p' },
 
+  { name: 'Object.setPrototypeOf(..., null)', src: `
+    var o = {parent: 'o'};
+    var q = Object.create(o);
+    Object.setPrototypeOf(q, null) === q &&
+        Object.getPrototypeOf(q);
```
NeilFraser: This fits on one line, no?

cpcallen: Yes.  Fixed.  I think the test might have at some point had more conditions.

PR #224, opened 2018-08-16
Make REPL more robust

cpcallen:
Prevent crashing when a syntax error is encountered (and don't bother printing stack trace in that case).

Do still crash if interpreter encounters internal error.

### DIFF
--- a/server/repl
+++ b/server/repl
@@ -48,10 +48,19 @@ const rl = readline.createInterface({
 
 rl.prompt();
 rl.on('line', function(line) {
-  const thread = intrp.createThreadForSrc(line).thread;
-  intrp.run();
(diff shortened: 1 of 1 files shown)


NeilFraser:
Lgtm


Sent (awkwardly) from an Android device.

On 16 Aug 2018 10:13, "Christopher Allen" <notifications@github.com> wrote:

@cpcallen <https://github.com/cpcallen> requested your review on:
NeilFraser/CodeCity#224 <https://github.com/NeilFraser/CodeCity/pull/224>
Make REPL more robust.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/NeilFraser/CodeCity/pull/224#event-1792493358>, or mute
the thread

PR #234, opened 2018-08-31
Throwing stuff around

NeilFraser:
* Throw *new* error (in accordance with the style guide).
* Throw string errors for user messages in code editor.
* Loop through property settings in code editor.


demo/core_13_$.utils.code.js
```diff
@@ -244,8 +244,8 @@ $.utils.code.rewriteForEval.unittest = function() {
       actual = SyntaxError;
     }
     if (actual !== cases[key][0]) {
-      throw Error('Eval Expression\n' +
-                  'Expected: ' + cases[key][0] + ' Actual: ' + actual);
+      throw new Error('Eval Expression\n' +
+                      'Expected: ' + cases[key][0] + ' Actual: ' + actual);
```
cpcallen: Comment: the style guide is not entirely clear about how the second line should be indented.  The most obvious interpretation is that the `'` should be aligned underneath the `w` of `throw`.  I personally prefer how you have it here.

NeilFraser: The old style guide offered an explicit choice:
https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting

PR #239, opened 2018-09-04
Pre demo

NeilFraser:
Final changes from me before tomorrow's demo.


demo/core_10_base.js
```diff
@@ -541,6 +562,31 @@ $.user.destroy.dobj = 'any';
...
+  var name = cmd.dobjstr.toLowerCase();
+  for (var key in $.userDatabase) {
+    var who = $.userDatabase[key];
+    if (who.name.toLowerCase() === name) {
+      if (who.location === this.location) {
+        user.narrate('' + who.name + ' is already here.');
+      }
+      if (user.location) {
+        user.location.narrate(user.name + ' vanishes into thin air.');
+      }
```
cpcallen: This should eventually be part of enterfunc / exitfunc—which I think I have actually implemented (the hook functions, not the announcements) in the version in `core/`.

demo/core_10_base.js
```diff
@@ -558,6 +558,31 @@ $.user.destroy.dobj = 'any';
 $.user.destroy.prep = 'none';
 $.user.destroy.iobj = 'none';
 
+$.user.join = function(cmd) {
+  var name = cmd.dobjstr.toLowerCase();
+  for (var key in $.userDatabase) {
+    var who = $.userDatabase[key];
+    if (who.name.toLowerCase() === name) {
+      if (who.location === this.location) {
+        user.narrate('' + who.name + ' is already here.');
```
cpcallen: Add a return here?

PR #245, opened 2018-10-19
Add some ES6 Math functions.

NeilFraser:
These are little-known functions that I was unaware of until today.

### DIFF
--- a/server/startup/es6.js
+++ b/server/startup/es6.js
@@ -58,8 +58,8 @@ var WeakMap = new 'WeakMap';
   var struct = [
     [Object, 'Object', ['is', 'setPrototypeOf'], []],
     [String, 'String', [], ['endsWith', 'includes', 'repeat', 'startsWith']],
-    [Number, 'Number', ['isFinite', 'isNaN', 'isSafeInteger'], []],
-    [Math, 'Math', ['sign', 'trunc'], []],
+    [Number, 'Number', ['isFinite', 'isInteger', 'isNaN', 'isSafeInteger'], []],
+    [Math, 'Math', ['acosh', 'asinh', 'atanh', 'cbrt', 'clz32', 'cosh', 'expm1', 'fround', 'hypot', 'imul', 'log10', 'log1p', 'log2', 'sign', 'sinh', 'tanh', 'trunc'], []],
     [WeakMap, 'WeakMap', [], ['delete', 'get', 'has', 'set']],
   ];
   for (var i = 0; i < struct.length; i++) {
@@ -216,5 +216,11 @@ Object.defineProperty(Number, 'MAX_SAFE_INTEGER',
     {configurable: false,
      enumerable: false,
      writable: false,
-     // Fortunately 2**53 is also safe as long as you dont' increment it!:
+     // Fortunately 2**53 is also safe as long as you don't increment it!:
      value: Math.pow(2, 53) - 1 });
+
+Object.defineProperty(Number, 'MIN_SAFE_INTEGER',
+    {configurable: false,
+     enumerable: false,
+     writable: false,
+     value: -Number.MAX_SAFE_INTEGER});
--- a/server/startup/es6.js
+++ b/server/startup/es6.js
@@ -59,7 +59,9 @@ var WeakMap = new 'WeakMap';
     [Object, 'Object', ['is', 'setPrototypeOf'], []],
     [String, 'String', [], ['endsWith', 'includes', 'repeat', 'startsWith']],
     [Number, 'Number', ['isFinite', 'isInteger', 'isNaN', 'isSafeInteger'], []],
-    [Math, 'Math', ['acosh', 'asinh', 'atanh', 'cbrt', 'clz32', 'cosh', 'expm1', 'fround', 'hypot', 'imul', 'log10', 'log1p', 'log2', 'sign', 'sinh', 'tanh', 'trunc'], []],
+    [Math, 'Math', ['acosh', 'asinh', 'atanh', 'cbrt', 'clz32', 'cosh', 'expm1',
+                    'fround', 'hypot', 'imul', 'log10', 'log1p', 'log2', 'sign',
+                    'sinh', 'tanh', 'trunc'], []],
     [WeakMap, 'WeakMap', [], ['delete', 'get', 'has', 'set']],
   ];
   for (var i = 0; i < struct.length; i++) {
@@ -92,11 +94,13 @@ var WeakMap = new 'WeakMap';
 
 // Polyfill copied from:
 // developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/find
-Object.defineProperty(Array.prototype, 'find', {value: function(callback/*, thisArg*/) {
+Object.defineProperty(Array.prototype, 'find', {value:
+    function(callback/*, thisArg*/) {
   if (this === null || this === undefined) {
     throw new TypeError('Array.prototype.find called on ' + this);
   } else if (typeof callback !== 'function') {
-    throw new TypeError('callback is type ' + typeof callback + ', not type function');
+    throw new TypeError('callback is type ' + typeof callback +
+                        ', not type function');
   }
   var o = Object(this);
   var len = o.length >>> 0;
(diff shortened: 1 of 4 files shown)


server/tests/db/test_01_es6.js
```diff
@@ -307,10 +307,21 @@ tests.NumberIsNaN = function() {
...
+  console.assert(!Number.isInteger(3.1), 'Number.isInteger 3.1');
+  console.assert(Number.isInteger(3.0), 'Number.isInteger 3.0');
+};
+
 tests.NumberIsSafeInteger = function() {
   console.assert(Number.isSafeInteger(3), 'Number.isSafeInteger 3');
-  console.assert(!Number.isSafeInteger(Math.pow(2, 53)), 'Number.isSafeInteger 2^53');
-  console.assert(Number.isSafeInteger(Math.pow(2, 53) - 1), 'Number.isSafeInteger 2^53-1');
+  console.assert(!Number.isSafeInteger(Math.pow(2, 53)), 'Number.isSafeInteger 2**53');
+  console.assert(Number.isSafeInteger(Math.pow(2, 53) - 1), 'Number.isSafeInteger 2**53-1');
```
cpcallen: Line length (two lines here, and ~14 more below).

server/startup/es6.js
```diff
@@ -58,8 +58,8 @@ var WeakMap = new 'WeakMap';
   var struct = [
     [Object, 'Object', ['is', 'setPrototypeOf'], []],
     [String, 'String', [], ['endsWith', 'includes', 'repeat', 'startsWith']],
-    [Number, 'Number', ['isFinite', 'isNaN', 'isSafeInteger'], []],
-    [Math, 'Math', ['sign', 'trunc'], []],
+    [Number, 'Number', ['isFinite', 'isInteger', 'isNaN', 'isSafeInteger'], []],
+    [Math, 'Math', ['acosh', 'asinh', 'atanh', 'cbrt', 'clz32', 'cosh', 'expm1', 'fround', 'hypot', 'imul', 'log10', 'log1p', 'log2', 'sign', 'sinh', 'tanh', 'trunc'], []],
```
cpcallen: We ought to fix the line length issues in the startup files, but perhaps in a separate PR.

server/tests/db/test_01_es6.js
```diff
@@ -327,6 +338,132 @@ tests.NumberMaxSafeInteger = function() {
       'Number.isSafeInteger(Number.MAX_SAFE_INTEGER + 1);');
 };
 
+tests.NumberMinSafeInteger = function() {
+  console.assert(Number.MIN_SAFE_INTEGER - 1 === -Math.pow(2, 53),
+      'Number.MIN_SAFE_INTEGER');
```
cpcallen: So, `-2**53` it isn't actually [safe according to the definition](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger), and this has the consequence that if `Number.MIN_SAFE_INTEGER` was mis-defined as -2**53 then this test would still pass.

server/tests/db/test_01_es6.js
```diff
@@ -327,6 +338,132 @@ tests.NumberMaxSafeInteger = function() {
...
+  console.assert(Number.isSafeInteger(Number.MIN_SAFE_INTEGER),
+      'Number.isSafeInteger(Number.MIN_SAFE_INTEGER);');
+  console.assert(!Number.isSafeInteger(Number.MIN_SAFE_INTEGER - 1),
+      'Number.isSafeInteger(Number.MIN_SAFE_INTEGER - 1);');
```
cpcallen: (But this test is O.K., because a "safe" integer can't be the result of rounding!)

server/tests/db/test_01_es6.js
```diff
@@ -327,6 +338,132 @@ tests.NumberMaxSafeInteger = function() {
...
+tests.MathCbrt = function() {
+  console.assert(Number.isNaN(Math.cbrt(NaN)), 'Math.cbrt NaN');
+  console.assert(Math.cbrt(-1) === -1, 'Math.cbrt -1');
+  console.assert(Object.is(Math.cbrt(-0), -0), 'Math.cbrt -0');
+  console.assert(Math.cbrt(-Infinity) === -Infinity, 'Math.cbrt -Infinity');
+  console.assert(Object.is(Math.cbrt(0), 0), 'Math.cbrt 0');
+  console.assert(Math.cbrt(1) === 1, 'Math.cbrt 1');
+  console.assert(Math.cbrt(Infinity) === Infinity, 'Math.cbrt Infinity');
+  console.assert(Math.cbrt(null) === 0, 'Math.cbrt null');
+  console.assert(Math.cbrt(2).toPrecision(10) === '1.259921050', 'Math.cbrt 2');
```
cpcallen: Maybe start with a really trivial case, like `Math.cbrt(27) === 3` or the like, in part just to make it obvious what the "cbrt" function is.

server/tests/db/test_01_es6.js
```diff
@@ -327,6 +338,132 @@ tests.NumberMaxSafeInteger = function() {
...
+  console.assert(Math.fround(1.5) === 1.5, 'Math.fround 1.5');
+  console.assert(Math.fround(1.337).toPrecision(10) === '1.337000012', 'Math.fround 1.337');
+  console.assert(Math.fround(Math.pow(2, 150)) === Infinity, 'Math.fround 2**150');
+  console.assert(Number.isNaN(Math.fround('abc')), 'Math.fround "abc"');
+  console.assert(Number.isNaN(Math.fround(NaN)), 'Math.fround NaN');
+};
+
+tests.MathHypot = function() {
+  console.assert(Math.hypot(3, 4) === 5, 'Math.hypot 3, 4');
+  console.assert(Math.hypot(3, 4, 5).toPrecision(10) === '7.071067812', 'Math.hypot 3, 4, 5');
```
cpcallen: Would it be useful to note that both (3, 4, 5) and (5, 12, 13) are Pythagorean triples?  (And of course that sqrt(x^2 + y^2 + z^2) = sqrt(sqrt(x^2 + y^2)^2 + z^2).)

server/tests/db/test_01_es6.js
```diff
@@ -327,6 +338,132 @@ tests.NumberMaxSafeInteger = function() {
...
+  console.assert(Math.hypot(3, 4, '5').toPrecision(10) === '7.071067812', 'Math.hypot 3, 4, "5"');
+  console.assert(Math.hypot(-3) === 3, 'Math.hypot -3');
+};
+
+tests.MathImul = function() {
+  console.assert(Math.imul(2, 4) === 8, 'Math.imul 2, 4');
+  console.assert(Math.imul(-1, 8) === -8, 'Math.imul -1, 8');
+  console.assert(Math.imul(-2, -2) === 4, 'Math.imul -2, -2');
+  console.assert(Math.imul(0xffffffff, 5) === -5, 'Math.imul 0xffffffff, 5');
+  console.assert(Math.imul(0xfffffffe, 5) === -10, 'Math.imul 0xfffffffe, 5');
```
cpcallen: I think it would be worth adding at least one test with non-integer inputs.

NeilFraser:
> Out of interest: how did these come to light?

While working on logarithmic backups I needed to compute an nth root.  While Googling, I found a bunch of rarely used Math functions that I hadn't seen before.  These functions are completely absent from most ES6 documentation, such as http://es6-features.org

server/tests/db/test_01_es6.js
NeilFraser: The Writely team (and others at Google) agreed that line length didn't apply to unit tests.  In general I think that makes things more readable.  As a specific example, imagine what this file would look like with 80 columns:
https://github.com/google/diff-match-patch/blob/master/javascript/tests/diff_match_patch_test.js#L406

Every year or so some open source contributor attempts to line wrap that file (or its cousins in other languages) and the result is nasty.

server/startup/es6.js
NeilFraser: Fixed.

server/tests/db/test_01_es6.js
NeilFraser: Changed to:
`Number.MIN_SAFE_INTEGER === -(Math.pow(2, 53) - 1)`

server/tests/db/test_01_es6.js
cpcallen: Well, tastes in aesthetics vary, but in this case I'd probably indent some of those tests like this:
```JS
  assertEquivalent(
      [[DIFF_DELETE, 'Apple'],
       [DIFF_INSERT, 'Banana'],
       [DIFF_EQUAL, 's are a'],
       [DIFF_INSERT, 'lso'],
       [DIFF_EQUAL, ' fruit.']],
      dmp.diff_main('Apples are a fruit.',
                    'Bananas are also fruit.', false));
```

(Though I still think it looks weird putting the expected output first.  Cursed yoda conditionals.)

In any case, `test_01_es6.js` is otherwise (mostly) in compliance with the JS style guide,  which enumerates the allowable exceptions to the 80-column limit—none of which appear to apply here.

PR #248, opened 2018-11-19
Production updates

cpcallen:
* Config files changes for running 'demo' core in prod with logarithmic deletion of backups and max 2GB of space for checkpoints.

* Treat checkpointMaxDirectorySize as specified in megabytes instead of bytes.

* Fix bug in computation of ideal times, where they would be spaced out between present time and most recent backup (rather than between present time and oldest backup).

### DIFF
--- a/server/codecity.js
+++ b/server/codecity.js
@@ -60,7 +60,7 @@ CodeCity.startup = function(configFile) {
     process.exit(1);
   }
   // Find the most recent database file.
-  var checkpoint = CodeCity.allCheckpoints().pop();
+  var checkpoint = CodeCity.allCheckpoints()[0];
   // Load the interpreter.
   CodeCity.interpreter = new Interpreter({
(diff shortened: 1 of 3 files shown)


server/codecity.js
```diff
@@ -148,7 +148,7 @@ CodeCity.parseJson = function(text) {
 CodeCity.allCheckpoints = function() {
   var files = fs.readdirSync(CodeCity.databaseDirectory);
   files = files.filter((file) => CodeCity.allCheckpoints.regexp_.test(file));
-  files.sort();
+  files.sort().reverse();
```
NeilFraser: Line 63: 
var checkpoint = CodeCity.allCheckpoints().pop();
Check if this needs to be [0] instead of pop.

cpcallen: (It now is.)

PR #249, opened 2018-11-21
Fix backup decay interval calculation

cpcallen:
There were some issues with how we were computing backup decay intervals.  This makes the following changes:

- idealTimes are computed based on a linear term (interval * n) plus an exponential one (decayRate ** n) - rather than just the latter - to avoid the problem of wanting backups closer together than the minimum interval.

- The decay rate is computed with an adjustment to take this linear factor into account, plus to be in units of interval, rather than ms.

Previous to these changes, the computation of the idealTimes resulted in idealTimes[0] being approximately 1 ms, idealTimes[1] being 1.x ms, etc.; in prod, where we presently keep 36 backups, this resulted in the server desiring that the 13 most recent backups all be less than 1s old.

### DIFF
--- a/server/codecity.js
+++ b/server/codecity.js
@@ -101,6 +101,7 @@ CodeCity.startup = function(configFile) {
   // Checkpoint at regular intervals.
   // TODO: Let the interval be configurable from the database.
   var interval = CodeCity.config.checkpointInterval || 600;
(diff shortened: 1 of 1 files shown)


NeilFraser:
The resulting graph looks *very* nice.  LGTM.
Apologies for the late review.

PR #250, opened 2019-01-08
Support {proto} and {owner} in Selector

cpcallen:
Also supports `^` as a synonym for `{proto}`.


cpcallen:
Not urgent - just a little bit of the work related to that which I've been doing on `Dumper` that can be reviewed independently.  Probably no reason to review individual commits except for amusement value.

cpcallen:
Verbal LGTM from @NeilFraser.

PR #251, opened 2019-01-09
Miscellaneous fixes

cpcallen:
There were a bunch of compilation warnings and test failures that had crept in on master.  These commits fix them.


cpcallen:
Verbal LGTM from @NeilFraser

PR #252, opened 2019-01-09
Impement stack size limit

cpcallen:
This limit is enforced only when a (native or user) Function is called, because we don't want to pester user with stack errors just because nested statements/expressions have taken them slightly over the limit.

RangeError is throw, for compatibility with Node.js - though we may want to implement a custom subclass of RangeError to make it easier for user code to detect this particular kind of error.


cpcallen:
Verbal LTGM from @NeilFraser

PR #253, opened 2019-01-14
Miscellaneous preparation for time limits

cpcallen:
Various miscellaneous changes made while preparing to implement thread run-time limits, but not directly related to that task.


cpcallen:
Verbal LGTM from @NeilFraser

PR #254, opened 2019-01-17
Implement time limit

cpcallen:
It's not exactly polished, and as discussed we probably want a (more restrictive) tick limit too, but these commits implement a 100ms time limit on all code run by the web and telnet servers, which can be adjusted by un-listening and then re-listening with a different timeLimit value.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -140,11 +140,13 @@ Interpreter.prototype.now = function() {
  * @param {!Interpreter.State} state Initial state
  * @param {number=} runAt Time at which thread should begin execution
  *     (default: now).
+ * @param {number=} timeLimit Maximum runtime without suspending (in ms).
  * @return {!Interpreter.prototype.Thread} Userland Thread object.
  */
-Interpreter.prototype.createThread = function(owner, state, runAt) {
+Interpreter.prototype.createThread = function(owner, state, runAt, timeLimit) {
   var id = this.threads_.length;
-  var thread = new Interpreter.Thread(id, state, runAt || this.now());
+  var thread =
+      new Interpreter.Thread(id, state, runAt || this.now(), timeLimit);
   this.threads_[this.threads_.length] = thread;
   this.go_();
   return new this.Thread(thread, owner);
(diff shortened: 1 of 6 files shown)


cpcallen:
I had intended to send this yesterday, but it turned out there were… bugs.

There are now… fewer bugs.

server/interpreter.js
```diff
@@ -2213,6 +2227,61 @@ Interpreter.prototype.initThread_ = function() {
...
+
+  // Properties of the Thread prototype object.
+  /**
+   * Decorator to add standard permission and type checks for Thread
+   * prototype methods.
+   * @param {!Interpreter.NativeCallImpl} func Function to decorate.
+   * @return {!Interpreter.NativeCallImpl} The decorated function.)
+   */
+  var withChecks = function(func) {
+    name = (name === undefined ? func.name : name);
```
NeilFraser: Elsewhere in the code we typically do:
`  x = (a === b) ? c : d;`
Not that it matters.

server/interpreter.js
```diff
@@ -2213,6 +2227,61 @@ Interpreter.prototype.initThread_ = function() {
...
+    id: 'Thread.prototype.setTimeLimit', length: 1,
+    /** @type {!Interpreter.NativeCallImpl} */
+    call: withChecks(function setTimeLimit(
+        intrp, thread, state, thisVal, args) {
+      var limit = args[0];
+      var perms = state.scope.perms;
+      var old = thisVal.thread.timeLimit || Number.MAX_VALUE;
+      if (typeof limit !== 'number' || Number.isNaN(limit)) {
+        throw new intrp.Error(perms, intrp.TYPE_ERROR,
+        'new limit must be a number (and not NaN)');
```
NeilFraser: Indent this line.

server/interpreter.js
```diff
@@ -2863,6 +2935,22 @@ Interpreter.prototype.setValue = function(ref, value, perms) {
   }
 };
 
+/**
+ * Check to see if the current thread has run too long.  Called at the
+ * top of loops and before making function calls.
+ * @param {!Interpreter.Owner} perms Perm to use to create Error object.
+ * @param {!Array<Interpreter.State>=} stack Current State stack.  If
+ *   supplied, it will be popped to remove top item (e.g.: Call state)
+ *   from stack trace.
```
NeilFraser: Add @ private

server/tests/testcases.js
```diff
@@ -2711,6 +2711,42 @@ module.exports = [
...
+  { name: 'Thread.prototype.setTimeLimit(...)', src: `
+    Thread.current().setTimeLimit(1000);
+    Thread.current().getTimeLimit();
+    `,
+    expected: 1000 },
+
+  // Check invalid time limits are rejected.
+  { name: 'Thread.prototype.setTimeLimit(/* invalid value */) throws', src: `
+    Thread.current().setTimeLimit(1000);
+    var invalid = [0, 1001, NaN, "foo", true, {}];
```
NeilFraser: Reason for not single quoting?

cpcallen: I think because `testcases.js` originated with golang implementation, which did not support single-quoted strings, and I've been inadvertently cargo-culting its style since then…

PR #263, opened 2019-02-13
CodeMirror upgrades and JSSP editor.

NeilFraser:


static/code/editor.js
```diff
@@ -953,7 +953,90 @@ Code.functionEditor.focus = function(userAction) {
...
+ * Create the DOM for this editor.
+ * @param {!Element} container DOM should be appended to this containing div.
+ */
+Code.jsspEditor.createDom = function(container) {
+  container.innerHTML = `
+<style>
+#jsspEditor {
+  position: absolute;
+  top: 60px;
+  bottom: 45px;
```
cpcallen: Might it not be cleaner to put editor styling in a .ass file, rather than having it inline in the .js files?

I guess if there was just one .css file for all the editors we'd load some CSS that wasn't always needed, while separating it out into a .css fie per editor and loading it dynamically would involve doing more fetches, but those do not seem like big penalties.

static/code/editor.js
```diff
@@ -835,6 +835,8 @@ Code.functionEditor.createDom = function(container) {
     tabSize: 2,
     undoDepth: 1024,
     lineNumbers: true,
+    continueComments: 'Enter',
```
cpcallen: 1. Is this the same as `continueComments: true`, or is `Enter` not the same as the key that on my Mac is named "return"—i.e., enter?
1. If so, isn't this going to be kind of annoying?
1. If so, might we want to consider additionally specifying `continueLineComment: false`?

static/code/editor.html
```diff
@@ -9,6 +9,8 @@
 
     <link rel="stylesheet" href="/static/CodeMirror/lib/codemirror.css">
     <script src="/static/CodeMirror/lib/codemirror.js"></script>
+    <script src="/static/CodeMirror/addon/edit/matchbrackets.js"></script>
```
cpcallen: Hooray!

cpcallen:
By the way, does the CodeMirror update fix bug #264?

cpcallen:
Don't forget we also need to update paths in `$.www.editor`…!

static/code/editor.js
NeilFraser: Fixed:  `continueComments: {continueLineComment: false}`

PR #269, opened 2019-02-21
Add Apache config template

NeilFraser:
This may gradually become obsolete, but it will always be a good starting point.

### DIFF
--- a/etc/apache.conf
+++ b/etc/apache.conf
@@ -1,0 +1,60 @@
+# Template for Apache configuration.
+
+<VirtualHost *:443>
+        # '.academy' should redirect to '.world'.
+        ServerName XXXXX.codecity.academy
+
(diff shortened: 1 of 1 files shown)


NeilFraser:
> I note that, compared to the nginx config, this doesn't serve on port 80 at all, whereas the nginx config redirects any listed domain port 80 to a canonical domain port 443.

Added.  It had been unintentionally omitted since my server already had such a redirect programmed in another config file.

> Also: does this issue actual redirects to .world from .academy etc., or just internally rewrite URL before serving?

The "R" flag on the Apache RewriteRule lines "forces an external redirect".

cpcallen:
Very nice.  We should probably fix the nginx config to also redirect HTTPS connections to the server's canonical name.

PR #270, opened 2019-02-22
Compare round-tripped data for changes.

NeilFraser:
Clicking on an incompatible editor should not nullify the data.

### DIFF
--- a/static/code/editor.js
+++ b/static/code/editor.js
@@ -446,7 +446,7 @@ Code.Editor.setSourceToAllEditors = function(src) {
   for (var editor of Code.Editor.editors) {
     editor.setSource(src);
     // Round-trip version of the source.
-    editor.lastSavedSource = editor.getSource();
+    editor.unmodifiedSource = editor.getSource();
   }
 };
 
@@ -650,7 +650,7 @@ Code.GenericEditor = function(name) {
    * Plain text representation of this editor's contents as of load or last save.
    * @type {?string}
    */
-  this.lastSavedSource = null;
+  this.unmodifiedSource = null;
 
   // Register this editor.
   Code.Editor.editors.push(this);
@@ -686,7 +686,7 @@ Code.GenericEditor.prototype.setSource = function(source) {
  * @return {boolean} True if work is saved.
  */
 Code.GenericEditor.prototype.isSaved = function() {
-  return this.getSource() === this.lastSavedSource;
+  return this.getSource() === this.unmodifiedSource;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


NeilFraser:
I ended up redesigning the system from scratch on paper, then working through the existing code to implement the diff.  Turned out that there was just one parameter that needed deleting.

static/code/editor.js
```diff
@@ -440,15 +440,13 @@ Code.Editor.mostConfidentEditor = function() {
...
-Code.Editor.setSourceToAllEditors = function(src, isSaved) {
+Code.Editor.setSourceToAllEditors = function(src) {
   Code.Editor.uncreatedEditorSource = src;
   for (var editor of Code.Editor.editors) {
     editor.setSource(src);
-    if (isSaved) {
-      editor.lastSavedSource = editor.getSource();
-    }
+    // Round-trip version of the source.
+    editor.lastSavedSource = editor.getSource();
```
cpcallen: O.K., I think I understand how this works, but maybe `lastSavedSource` should be renamed?  It seems to me it's *not* the last saved source, but rather the just the unmodified-by-the-user source.

NeilFraser: Renamed to `unmodifiedSource`.

NeilFraser:
It seems to behave exactly as one would want, with no trivial changes appearing unexpectedly.

Note that the grey save button is still live, and can still perform a save.  This allows one to force a trivial change that the editor doesn't deem worthy.  It also allows one to force a save and revert a value that was edited in another window (something I did all the time in Moo Canada, open two copies of the editor, and use one to execute reverts).

There is one slight wrinkle:  If one loads a function, and switches to the SVG editor, one gets a blank canvas as expected.  If one doesn't edit the SVG, but instead returns to the function editor, the original function is presented.  Regardless of tab switching, if the user hasn't touched anything the save button will never go blue.  This is all good.  However, if in the SVG editor one presses the grey save button, it will not save the blank SVG that is currently being presented, but rather resave the function.

This is a bit unexpected.  But after thinking about it for some time, I am of the opinion that this is the best available behaviour.  The reasons are:  a) There's rarely a reason to overwrite a value of a different type with a blank JSSP/SVG/function/etc.  b) The presence of a grey save button should consistently indicate that saving won't do anything (or at least not much).

cpcallen:
That's totally satisfactory.  Go ahead and commit this.  I can push it to prod if you like.

cpcallen:
Are you still tracking down some tab-clickhappiness weirdness?  It would be very nice to land this ASAP.  If you think it's not likely to be a major show-stopper, can it be dealt with later?

Issue #268, opened 2019-02-18
Unable to log in "Google Authentication fail: Error: invalid_grant"


cpcallen:
I am unable to log in using my @gmail.com account; I get the message "Google Authentication fail: Error: invalid_grant" at `https://google.codecity.world/login?code=[CENSORED]&scope=openid+email+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email#`.

cpcallen:
O.K., I am able to reproduce this 100% of the time on my local machine, using an incognito window to log in with either my corp or personal account.

cpcallen:
More info: after getting "Invalid Grant" message, if you go back to the login page and click the login button again it works fine.

cpcallen:
The first time we hit the login server, the URL is just `/login`.  After user clicks login, we hit it again with a URL that looks like:
```
/login?code=4%2F_gChz... [censored] ...Se98&scope=openid+email+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email
```
This provokes the `invalid_grant` response, but if the user goes back to the login page and clicks the button again, they hit the login server with a URL that looks like:
```
/login?code=4/_gCMN... [censored] ...rM7g&scope=openid+email+https://www.googleapis.com/auth/userinfo.email
```
… and that one works.

cpcallen:
O.K., confirmed: the problem is that `oauth2Client.getToken` doesn't like being passed the code URI encoded.  (It also doesn't *need* to get passed the rest of the query parameters, though seem to ignore them happily enough.)

PR #273, opened 2019-02-27
Improvements to loginServer

cpcallen:
Fix bug #268 and make various code-quality improvements to loginServer.  Changes probably most easily reviewed individually.

### DIFF
--- a/login/loginServer.js
+++ b/login/loginServer.js
@@ -32,6 +32,9 @@ var google = require('googleapis');
 var http = require('http');
 var URL = require('url').URL;
 
+// Configuration constants.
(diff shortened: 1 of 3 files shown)


cpcallen:
> I thought that in general we were restricting our server-side JS to match what Code City supported so that as developers we had a consistent programming experience.

I've reworked this to remove ES6-isms, but to keep the rest of the refactoring I did.  See commit 
0cd33f4, which is the only substantial change.

> Also, should package-lock.json be checked in at all?

[Yes](https://docs.npmjs.com/files/package-lock.json).

PR #410, opened 2020-07-23
Restore editor login page, fix MobWrite

NeilFraser:
Setting window.closeMe no longer works since cross orgin frame ‘login.’ to code.’


NeilFraser:
Filed bug with Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1109065

mobwrite/mobwrite_server.py
```diff
@@ -227,7 +227,8 @@ def do_POST(self):
     data = data[div + 2:]
     data = urllib.unquote(data)
     self.send_response(200)
-    self.send_header('Content-type','text/plain')
+    self.send_header('Content-type', 'text/plain')
+    self.send_header('Access-Control-Allow-Origin', '*')
```
cpcallen, line 231: Ought this not to be limited to `example.codecity.world` and subdomains thereof?

cpcallen:
@googlebot I consent.

cpcallen:
So, modulo the my one comment on `mobwrite/mobwrite_server.py` above (and any comment you might like to make about check-in of `core/`) I think this PR is ready to go.  However: when I tested this locally it appeared that login was broken—i.e., if I delete my ID cookie and then try to log in via the login button on `/` or via the code editor pop-up at `/code/login`, I just keep getting redirected back to the same page.

Probably be good to fix that before committing this.

cpcallen:
> > * Fix some over-zealous application of String(this) to get physical
> >   objects' names (e.g., in command menus).
> 
> BTW, this is fixing a bug where the genetic mice had command menus of:
> 
> look Squeaky (Neil/Neil/Neil)
> …

I didn't guess this particular cause, but I figured it would be something like that.  In any case, I'd already backed out a few such applications of `String()` when creating memos, because for sure we should be using the (unique) `.name` and not any kind of title when providing the client with identifiers for rooms, users, etc.  It makes sense that the same should apply to commands, although here we could also choose an alias, if a nice short one was available.  (We'd need to do a disambiguation check, though, since aliases are not unique, so that starts to get a bit complicated.)

PR #412, opened 2020-08-13
Terminate connect and mobwrite when logged out

NeilFraser:


connect/connectServer
```diff
@@ -268,6 +268,13 @@ function handleRequest(request, response) {
       }
     });
     request.on('end', function() {
+      // No ID cookie, the user has logged out.
+      if (!/(^|;)\s*ID=\w/.test(request.headers.cookie)) {
```
cpcallen, line 272: I'm glad of the comment.  Could the check itself be made any easier to understand?  It would be lovely if it could read something like `if(!cookies.has('ID')) {`… (presuming that `request.headers.cookie` has been split out into a Map named `cookies`).

mobwrite/mobwrite_server.py
```diff
@@ -211,33 +212,43 @@ def do_POST(self):
           (self.client_address[0], connection_origin))
     mobwrite_core.LOG.info("Connection accepted from " + self.client_address[0])
 
+    required_cookie = mobwrite_core.CFG.get("REQUIRED_COOKIE", "")
+    if required_cookie and (('Cookie' not in self.headers) or
+        (not re.search(r'(^|;)\s*%s=\w' % required_cookie, self.headers['Cookie']))):
```
cpcallen, line 217: See comment above about connectServer's similar check.

PR #413, opened 2020-08-21
Better nginx config

cpcallen:
* Better handling for ACME HTTP-01 challenges.
* Factor out `proxy` config.
* Disable auto-retry of GET requests to CodeCity server.
* Documentation copyedit.

### DIFF
--- a/etc/cc-localhost.conf
+++ b/etc/cc-localhost.conf
@@ -27,8 +27,8 @@ server {
     # Static files.
     autoindex on;
     index index.html;
-    # Edit to be full path to CodeCity directory -
-    # e.g. /home/userid/src/CodeCity
+    # Edit to be full path to CodeCity directory.
+    # E.g. /home/userid/src/CodeCity
     root REPOSITORY;
   }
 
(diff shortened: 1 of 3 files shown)


cpcallen:
Ahem.

naman1236:
Sir I want to become software engineer
So what should I do after 10th ??

NeilFraser (approved):

etc/cc-localhost.conf
```diff
@@ -2,16 +2,25 @@
...
 # The easiest way to use this file is to leave it unedited and instead
-# start nginx using bin/nginx-dev, which will dynamically creat# suitable config files on the fly.
+# start nginx using bin/nginx-dev, which will dynamically create
+# suitable config files on the fly.
+
+# Configuration applying to all servers.
+error_page 502 503 504 =503 /static/503.html;
+
+proxy_set_header Host $http_host;
+proxy_next_upstream_tries 1;
```
cpcallen, line 12: Turns out this causes (or at least surfaces) an unanticipated problem; see #416.

PR #415, opened 2020-08-30
Reformatted for easier reading, updating link, better notes

Grigoletti2001:


cpcallen:
Hi Joseph,

Thanks for your contribution.  Unfortunately this file is a vendored copy of [SVG-edit](https://github.com/SVG-Edit/svgedit), so your changes would need to be submitted to that upstream instead—after which we *might*, *eventually*, update this copy.

Christopher

Grigoletti2001:
Thank you for the feedback.

PR #427, opened 2020-09-27
Add run on repl.it badge to README

JoeyCreator:
This pull request adds a  badge to the . This will allow users to easily run this repository in their browser, without having to set up an environment. You can learn more about Repl.it [here](https://repl.it).

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "bash"
+run = "echo hello world"
--- a/README.md
+++ b/README.md
@@ -5,3 +5,4 @@ education.  It offers a comic book inspired virtual world where programmers can
 write code collaboratively.
 
 A list of running Code City instances may be found at https://codecity.world/
+[![Run on Repl.it](https://repl.it/badge/github/JoeyCreator/CodeCity)](https://repl.it/github/JoeyCreator/CodeCity)


cpcallen:
Oh, this would be great—but I tried out the button link just now and didn't get anything resembling a functioning CodeCity instance as a result.

I'm not especially familiar with how [repl.it](https://repl.it) works, but if you can actually get this to work that would be fantastic.  A few notes to help you along the way:

* CodeCity generally depends on node.js, but it ought in principle to be able to run the CodeCity interpreter proper entirely in the browser.  However:
   * Most major browsers now support support [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef), but it will be necessary to update `server/iterable_weakmap.js` and `server/iterable_weakset.js` to use `FinalizationRegistry` instead of `FinalizationGroup` (noting that both the API and the name have changed somewhat).
   * It will also be necessary to provide alternatives to other Node functionality.  The few things we use from the node `util` package should be easy to provide browser alternatives for (I'm sure someone has already written polyfills for these), but it will probably be necessary to disable the networking builtins (`CC.listen` et al.) since there isn't really any way to do those things in a browser.  Except `CC.xhr`; that one will be easy to replace with `XmlHttpReqest`.
* If it's not actually running in a browser but on a repl.it VM, then [the setup instructions](https://github.com/google/CodeCity/blob/master/docs/setup.md) should provide most of what you need to know to be able to get it running.  You'll probably want to replace the existing Google OAuth login service with one connected to repl.it accounts, and I'm not sure how you will handle DNS, TLS certs and so on, but the rest should be straight forward.
* A simpler alternative might be to fire up the `server/repl`](https://github.com/google/CodeCity/blob/master/server/repl) script, which provides a CodeCity-flavoured-JS read/eval/print loop, albeit without checkpointing, etc.

JoeyCreator:
I try to add a link to the run on repl.it button but on the README.md file doesn't look good
Capture d’écran (44)

JoeyCreator:
I have to configure the run button and have enter command.

cpcallen:
I'm going to close this for now, but I would be delighted to re-open it if you can get Code City working on repl.it.

PR #428, opened 2020-09-28
Python 2 required for MobWrite

NeilFraser:
Python 3 throws "No module named 'thread'"


cpcallen (approved):
Looks good, but `etc/codecity-mobwrite` should probably be updated too—or maybe better still, the choice of python (major) version should be baked into a shebang line at the top of `mobwrite_server`.

cpcallen (approved):

PR #434, opened 2020-11-20
Simplify builtin Array.prototype.join implementation

cpcallen:
No need for a state machine if we just return our callee's return value.


NeilFraser (approved):

cpcallen:
This was merged instead of rebased.  I'm going to force-push to remove the merge commit from `master` branch.

PR #440, opened 2021-01-28
Add ‘null’ symbol to code editor

NeilFraser:
Both google.codecity and blockly.games have the corresponding database update.

### DIFF
--- a/static/code/objectPanel.js
+++ b/static/code/objectPanel.js
@@ -126,6 +126,7 @@ Code.ObjectPanel.TYPES = {
   'array': '[]',
   'boolean': '⏼',
   'function': '𝑓',
+  'null': '␀',
   'number': '#',
   'object': '{}',
   'string': '”',


static/code/objectPanel.js
```diff
@@ -126,6 +126,7 @@ Code.ObjectPanel.TYPES = {
   'array': '[]',
   'boolean': '⏼',
   'function': '𝑓',
+  'null': '␀',
```
cpcallen, line 129: Oh, nice.  I was going to suggest using "∅", but—provided it is _legible_—this is even more clear.

cpcallen (approved):

static/code/objectPanel.js
NeilFraser: I too have concerns on that front.  Chrome renders this character tiny with lots of surrounding whitespace.  Firefox is better.  And Android renders 'NUL' diagonally.  So much diversity!

Whatever, we can always change it.

cpcallen: > Whatever, we can always change it.

My thought exactly.

PR #442, opened 2021-01-28
loginServer updates, including ?after= custom post-login redirect target

cpcallen:
* Update to ES6 and async/await to make control flow easier to follow.
* Have close page served statically rather than by `loginServer`.
* Support `?after=` param to specify post-login redirect target
   * For now continue to support `?loginThenClose` (and default redirect to `connectServer`) for backwards compatibility; these can be removed in future when homepage and editor-login pages have been updated on all running instances.  ([These](https://code.google.codecity.world/?$.http%5B%22www.%22%5D%5B%22%2F%22%5D.www.source) [updates]($.http["code."]["/login"].www.source) have already been made on `google`, though they will have no effect until this PR is pushed to prod, and even then the effect will be no change.)
* Make URL-independent: does not need to know own loginUrl, and can therefore serve multiple URLs as required.


NeilFraser (approved):
Oof, that's a lot of ES6.  I thought we'd decided to only use language that the server could also understand (unless there were exceptional circumstances).

PR #443, opened 2021-02-04
Core update

cpcallen:
Pull the latest changes from the `google.codecity.world` database into `core/`.  See individual commits for details.


cpcallen:
Leaving this as a draft pending `$.Jssp` updates.

cpcallen:
OK, the big `$.Jssp` -> `$.jssp` transition has now fully landed, so I think this can now be merged, if you're happy with it.

There's a bunch of ongoing (not yet finished) work on routing in the web server:

- Creating a `$.servers.http.Host` class to represent a "virtual host"—root or subdomain—with various mostly-optional configuration properties like `.hostname`, `.scheme` (i.e., `http` or `http`), and `.pathToSubdomain` (the inverse of the previous `.subdomains` flag), as well as a `.subdomains` (null-proto object) map.  The `Host.prototype` object has default values for many of these and additionally has various methods.

- There's a directory of Host objects at `$.hosts`, which will replace `$.http` in due course.  (Not fussed about the name; we can rename this to `$.http` if you like.)  This is for convenient editor access.

- For handling/routing purposes, `$.servers.http.onRequest` uses an array `$.servers.http.hosts`, calling each `$.servers.http.hosts[n].handle(request, response)` until one of the Hosts accepts the request, handles it, and returns `true`.  In the default configuration, `$.servers.http.hosts` will contain just one entry—`$.hosts.root`—but by adding additional entries to this array you can serve an arbitrary number of virtual hosts.

NeilFraser (approved):

PR #464, opened 2021-03-19
Implement a login service backend

cpcallen:
Have loginServer send some of the data it got from the OAuth login to the database as a JSON string via a new server, `$.servers.login`, listening on port 7776, which will return a string to use as the ID cookie.

Also enable half-open sockets, so that `loginServer` (or indeed a web browser) can close its end of the connection as soon as it is done sending a query without preventing the responding server from sending data back.

### DIFF
--- a/login/loginServer
+++ b/login/loginServer
@@ -28,6 +28,7 @@ const forwardedParse = require('forwarded-parse');
 const fs = require('fs').promises;
 const {google} = require('googleapis');
 const http = require('http');
+const net = require('net');
 const {URL, format: urlFormat} = require('url');
 
 const oauth2Api = google.oauth2('v2');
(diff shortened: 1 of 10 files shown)


login/loginServer
```diff
@@ -56,10 +57,26 @@ const DEFAULT_CFG = {
...
-  emailRegexp: '.*'
+  emailRegexp: '.*',
+  // Port number for the login service backend.
+  backendPort: 7776,
+  /* List of fields, from data object returned by
+   * oauth2Api.userinfo.v2.me.get, to pass along in the request the
+   * login service backend.
+   *
+   * Available fieldnames, and the types and meanings of their values:
+   * - id: string - the user's OAuth (GAIA) ID as a numeric string.
```
NeilFraser, line 68: I think it's important to note here that this is salted/hashed.

NeilFraser (approved):

PR #470, opened 2021-03-29
Polyfill Object.getOwnPropertyDescriptors

cpcallen:


server/startup/es8.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
NeilFraser, line 3: 2021

NeilFraser (approved):

server/startup/es8.js
cpcallen: Fixed.

PR #475, opened 2021-04-19
Fix network-related test failures

cpcallen:
The goal of this PR is to clean up compile and test failures.  There are quite a few of the latter, so here we concentrate on the ones which were causing the tests to hang.  Specifically:

* Fix compile errors that had been caused by enabling `allowHalfOpen`.
* Fix test hangs due to enabling `allowHalfOpen`.
* Some additional tidying of network-related tests.
* Attempt to add two seemingly-innocuous tests of interaction between networking code and the `start`, `stop` and `pause` methods on `Interpreter`, but discover that there are some async-related "race conditions" (not actually indeterminate—just unexpected) that cause test crashes, revealing the present network code is even more fragile than previously believed.
* Substantially refactor `Interpreter.prototype.Server` to fix this, but realise that `Serializer` now mungs `Server.server_`.
* Take a diversion and make several improvements to `Serializer`.
* Finally check in the refactored `Server` code and the innocuous-looking tests that motivated it.

(More details can be found in the individual commit messages.)

### DIFF
--- a/server/serialize.js
+++ b/server/serialize.js
@@ -30,44 +30,98 @@ var Registry = require('./registry');
 
 var Serializer = {};
 
-/** !Set<!Object> Prototypes of types to be excluded from serialization. */
-Serializer.excludeTypes = new Set([net.Socket.prototype, net.Server.prototype]);
+/** 
+ * A record of a single serializalbe type, including a type tag, a
+ * constructor function to call when deserializing an instance, and an
+ * optional list of properties to exclude when serializing.
+ *
+ * @typedef {{
+ *   tag: string,
+ *   constructor: !Function,
+ *   prune: (!Array<string>|undefined),
+ * }}
+ */
+var TypeInfo;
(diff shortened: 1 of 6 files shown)


server/serialize.js
```diff
@@ -30,6 +30,107 @@ var Registry = require('./registry');
...
+ * @typedef {{
+ *   types: !Array<!TypeInfo>,
+ *   byTag: !Object<!TypeInfo>,
+ *   byProto: !Map<?Object,!TypeInfo>,
+ * }}
+ */
+var Config;
+
+/**
+ * Create a configuration object for serializing or desieralizing na
```
NeilFraser, line 59: na?

NeilFraser (approved):

server/serialize.js
cpcallen: Stupid butterfly keyboard.

PR #476, opened 2021-04-20
Have Object.setOwnerOf return the object

cpcallen:
This is to make it conform to the custom established by `Object.setPrototypeOf` and `Object.defineProperties`, and allows code like:

```JS
obj = Object.setOwnerOf(
          Object.setPrototypeOf(
              Object.defineProperties({
                key: 'value',
              }, {
                key: {enumerable: false},
              ),
              myProto),
          myOwner);
```

Not pretty - but pretty useful in some cases, such as in the editor.


NeilFraser (approved):

cpcallen:
Thanks!

PR #477, opened 2021-04-20
Update dependencies, including to Node.js v14

cpcallen:
Update Acorn, Closure Compiler to the latest available npm version, and update our code to run on Node.js v14, which finally includes `WeakRef` as standard.

Also fix some compile warnings, create some more and fix most of them, and fix some tests.

Left for a future PR are one compile warning and one test failure, both of which concern buggy code in `Dumper`.

### DIFF
--- a/server/externs/WeakRef.js
+++ b/server/externs/WeakRef.js
@@ -24,20 +24,6 @@
 
 /**
  * @constructor
- * @param {T} target
- * @template T
- */
-// TODO(cpcallen): Make T bounded to {!Object} once closure-compiler
-// supports bounded generic types.
-var WeakRef = function(target) {};
-
-/**
- * @return {T}
- */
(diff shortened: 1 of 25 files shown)


server/externs/WeakRef.js
```diff
@@ -16,52 +16,37 @@
  */
 
 /**
- * @fileoverview Closure Compiler externs for the new tc39 WeakRef and
- *     FinalizationGroup API https://github.com/tc39/proposal-weakrefs
+ * @fileoverview Closure Compiler externs for the new ES2020  WeakRef and
```
NeilFraser, line 19: Looks like an extra space here.

NeilFraser (approved):

server/externs/WeakRef.js
cpcallen: Fixed.

PR #478, opened 2021-04-21
Dumper code-generation improvements: use .write's var_args and .exprForCall_

cpcallen:
* Use `Dumper.prototype.write`'s ability to accept variable numbers of arguments to avoid sting concatenation where possible.
* Introduce `Dumper. prototype.exprForCall_` to simplify generation of calls to builtin functions.
* Fix test of `Dumper.prototype.dump` that was broken when code to dump `Server` `.timeLimit`s added in 91d3959  (PR #453)

### DIFF
--- a/server/dumper.js
+++ b/server/dumper.js
@@ -198,14 +198,14 @@ Dumper.prototype.dump = function() {
   // Dump all remaining bindings.
   this.global.dump(this);
 
-  // Dump listening sockets.
+  // Dump listening Servers.
   for (var key in this.intrp2.listeners_) {
     var port = Number(key);
-    var listener = this.intrp2.listeners_[port];
+    var server = this.intrp2.listeners_[port];
     this.write(this.exprForBuiltin_('CC.connectionListen'), '(',
-               String(port), ',',
-               this.exprFor_(listener.proto), ', ',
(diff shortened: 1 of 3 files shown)


cpcallen:
As we discussed when reviewing a previous Dumper PR.

server/dumper.js
```diff
@@ -279,19 +280,25 @@ Dumper.prototype.exprFor_ = function(value, ref, callable, funcName) {
...
   if (key) {
     var quoted = code.quote(key);
     return callable ? '(new ' + quoted + ')' : 'new ' + quoted;
   }
-  // New object.  Create and save referece for later use.
-  if (!ref) throw Error('Refusing to create non-referable object');
+
+  // Seems to be a new object.  Check it really doesn't exist already
+  // and that it will be referenceable.
+  if (objDumper.proto !== undefined) throw new Error('object already exists');
```
NeilFraser, line 299: "throw new Error" here, "throw Error" below.  Pick one.

NeilFraser (approved):

server/dumper.js
cpcallen: Style guide says to use "new".  I think

PR #480, opened 2021-04-23
Tidy up testcases.js

cpcallen:
Numerous mostly-reformatting changes to `server/tests/testcases.js`:

* Run `clang-format` on the file to fix various formatting sins.
* Remove extraneous blank lines between testcases.
* Reduce simple testcases to a single line (and put source on the same line as the `src:` tag).
* Remove `name`s from testcases where it is the same as `src` or where it does not contribute any useful information about the purpose of the test.
* Fix formatting of the testcase source code itself:
  * Consistently indent source code by +2 spaces compared to surrounding testcase.
  * Apply certain [styleguide](https://google.github.io/styleguide/jsguide.html) rules regarding whitespace, line breaks, etc.
* Make selective improvements to testcase naming and comments.
* Don't use table-driven tests (within a single testcase) unnecessarily.
  * There are still some cases of table-driven tests where we are checking for certain errors being thrown, which is harder to to concisely as a top-level testcase.

Also:

* Backport a doc fix for `regexps.identifier` in `server/code.js` from the core.
* Add two tests concerning setting anonymous FunctionExpression `.name` properties by property assignment.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+Language:        JavaScript
+BasedOnStyle:    Google
+ColumnLimit:     80
--- a/server/code.js
+++ b/server/code.js
@@ -69,9 +69,10 @@ regexps.string = new RegExp('(?:' + regexps.singleQuotedString.source + '|' +
 regexps.stringExact = new RegExp('^' + regexps.string.source + '$');
 
 /**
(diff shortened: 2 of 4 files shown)


cpcallen:
OK, it took a few tries to get everything committed the way I intended, but this is now ready for review.

NeilFraser (approved):

PR #486, opened 2022-02-03
Fix `Object.keys`, improve `Object.assign`

cpcallen:
* Fix the `Object.keys` builtin, which was previously returning the same thing as `Object.getOwnPropertyNames`—i.e., including non-enumerable properties.
* Improve the `Object.assign` polyfill:
  * Use `Object.keys` (now fixed!) to obtain list of enumerable own keys, and iterate over that rather than using a `for in` loop.  This avoids having to check whether the keys are own keys or inherited.
  * Make `Object.assign` non-enumerable.
* Add tests.

### DIFF
--- a/server/interpreter.js
+++ b/server/interpreter.js
@@ -693,7 +693,7 @@ Interpreter.prototype.initObject_ = function() {
           typeof value === 'string') {
         // No boxed primitives in Code City.
         throw new intrp.Error(state.scope.perms, intrp.TYPE_ERROR,
-            'Boxed primitives not supported.');
+            'boxed primitives not supported');
       } else if (value === undefined || value === null) {
         return new intrp.Object(state.scope.perms);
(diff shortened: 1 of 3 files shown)


cpcallen:
@NeilFraser: PTAL.

server/interpreter.js
```diff
@@ -693,7 +693,7 @@ Interpreter.prototype.initObject_ = function() {
           typeof value === 'string') {
         // No boxed primitives in Code City.
         throw new intrp.Error(state.scope.perms, intrp.TYPE_ERROR,
-            'Boxed primitives not supported.');
```
NeilFraser, line 696: In my experience, JS errors don't have trailing punctuation.

cpcallen: And indeed I removed it.

